PR #1, opened 2021-07-18
Various tweaks

mkst:
Thought I'd take a look at this project - I had to make a few tweaks to get it to build - so have incorporated those changes to this PR.

It's up to you, but I don't think you want to be checking the asm files into the repo as they are extracted by splat - but guess that is purely personal preference.

I am unsure why the function definitions have been lost for osViBlack, osViSwapBuffer, osInitialize, osCreateThread, osStartThread. I'll ask Ethan about it.


Drahsid:
Thanks for the PR. I never extracted the libultra functions that I had named, so I don't believe splat has anything to use (besides the symbols.txt) to know that they are named.

PR #3, opened 2021-07-31
More matches, ID's and some #define's

blackgamma7:


Drahsid:
It looks like the head you began this on is out of date, many of these symbols are not named the same as they are here. Could you pull and merge your changes?

Drahsid:
I removed the asm directory from the repo  history, which force closed this question (since the history changed,) merge with the current master and reopen the PR. If you get issues with this, try running `git filter-branch --tree-filter 'rm -rf asm' HEAD` and then force pushing to your fork, that should settle things.

PR #15, opened 2026-01-30
Decompile func_800486F4 (marina)

DodongoRex:
Decompiles func_800486F4 in src/marina.c.

- Function matches (checksum OK, first_diff.py: No differences)
- Marina segment already compiling as C on cleanup2024
- Adds required symbol declarations and address-only linker symbols


DodongoRex:
Confirmed matching on US1.
Built with NON_MATCHING=1 (checksum OK) and ran first_diff.py — no differences.
Adds the required symbol plumbing now that marina is C (including D_801373F0 / D_801373F1 and related address-only symbols).

DodongoRex:
Rebased onto reserve-src-49200. Updated C to reproduce the ROM’s sign-extension behavior exactly.
Checksum OK
first_diff.py: No differences
marina.c emits .text only (no .data, .rodata, or .bss)
Ready to merge

PR #16, opened 2026-01-31
Decomp func_80048C28 (marina)

DodongoRex:
Removes the old #pragma GLOBAL_ASM for the function and deletes the corresponding nonmatching ASM file.
Implements func_80048C28 in C, matching the behavior of the original MIPS.
Adds an extern u16 D_801373D8 to include/data_symbols.h to expose the global used by this function.


Drahsid:

PR #17, opened 2026-01-31
Decompile func_80048C94 (marina)

DodongoRex:
Function is a small leaf routine (index → int → float multiply → truncate back to int).
C output compiles cleanly and emits an instruction-for-instruction match against the original ASM.
Matching was verified via objdump comparison (same instruction order and control flow; only expected register allocation differences).


Drahsid:

PR #18, opened 2026-02-01
Decompile func_80048CE4 (marina)

DodongoRex:
Decompiles func_80048CE4 in src/marina.c (marina segment).
The function is a small classifier over the byte global D_800EF5E1. The C uses a straightforward early-return if chain to mirror the original control flow and load-width (lbu/unsigned byte semantics).


Drahsid:

PR #19, opened 2026-05-07
Add Cosine and unused float tables. add macros to use table for sine and cosine.

blackgamma7:
This game uses a LUT of the cosine function for the near entirety of its sine and cosine routines. This PR includes the table (along with an unused LUT next to it) as well as several macros based on observed uses of this LUT.


versions/us1/symbol_addrs_data.txt
Drahsid, line 1: `gUnusedFloatTable =0x800BDCD0 //size:0x800`
Missing the `;` at the end here

include/cosine.h
Drahsid, line 1: ARRAYLENGTH/END probably belong in the common header.

Drahsid (changes requested):
Looks good. Send a quick patch for the two things below and LGTM!
Also, if you're getting back into this and want to claim specific files or overlays, let me know!

PR #21, opened 2026-05-08
Letterbox: Create header, add function labels, rename script, split from lifebar script

blackgamma7:
Start documenting the game's "Letterbox" function, by creating a descriptive header, adding the relevant symbols, and renaming the relevant script, which is separated from the lifebar's file.


Drahsid:
Are you planning on matching the functions in letterbox.c/lifebox.c?

blackgamma7:
> Are you planning on matching the functions in letterbox.c/lifebox.c?

Wanted to document letterbox script and better split it.

Tried to match 3 funcs. succeeded with one.

src/letterbox.c
Drahsid, line 1: letterbox.c oddly appears to be -O1.
I got Gfx_DrawLetterboxStandard pretty close: https://decomp.me/scratch/JNu1y
Not sure what caused this codegen with the 0/255 color, but it's also the key to Gfx_DrawLetterbox. It doesn't make much sense to divide by 255 in the GPACK_RGBA5551 macro.

blackgamma7: added more accurate code to PR.

PR #22, opened 2026-05-12
Build improvements: support mips64-elf, clang, verbose

queueRAM:
- Add support for mips64-elf and mips64-linux-gnu binutils, autodetect
- Add support for using clang as `cpp`, autodetect
- Add `--verbose` flag used for clean and passed to ninja

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -84,6 +84,15 @@ def get_version_num():
     elif GAME_VERSION == "eu":
         return 3
 
+def preferred_tool(programs: List[str]) -> str:
+    """determine which if any executable is available in PATH"""
+    for prog in programs:
+        for path in os.environ.get("PATH", "").split(os.pathsep):
+            fpath = Path(path) / prog
+            if fpath.is_file and os.access(fpath, os.X_OK):
+                return prog
+    return ""
+
 YAML_FILE = f"versions/{GAME_VERSION}/{BASENAME}.yaml"
 LD_PATH = f"versions/{GAME_VERSION}/{BASENAME}.ld"
 MAP_PATH = f"build/{BASENAME}.map"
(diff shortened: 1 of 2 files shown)


Drahsid:
I was able to confirm that this isn't breaking for mips-linux-gnu-* and I'm assuming you tested with the other available toolchains.
My only concern before merging is that you mentioned a `set_o32abi_bit.py` script to fix an issue you had in discord. Is that not needed?

configure.py
```diff
@@ -400,7 +414,7 @@ def build(
     ninja.rule(
         "as",
         description="as $in",
-        command=f"{CROSS_CPP} {COMMON_INCLUDES} -DGAME_VERSION={get_version_num()} $in -o - | iconv -t EUC-JP | {CROSS_AS} -G0 {COMMON_INCLUDES} -EB -mtune=vr4300 -march=vr4300 -o $out",
+        command=f"{CPP} {CPP_FLAGS} {COMMON_INCLUDES} -D_LANGUAGE_ASSEMBLY -DGAME_VERSION={get_version_num()} $in -o - | iconv -t EUC-JP | {CROSS_AS} -G0 {COMMON_INCLUDES} -EB -mtune=vr4300 -march=vr4300 -o $out && {O32_TOOL} $out",
```
queueRAM, line 417: This was the change needed to fix the linker errors with mips64-elf biuntils. This runs `O32_TOOL` for all assembly files, not just libultra.

queueRAM:
> I was able to confirm that this isn't breaking for mips-linux-gnu-* and I'm assuming you tested with the other available toolchains.

I tested mips64-elf on macOS arm64 and mips-linux-gnu under debian aarch64. I'll test some more hosts as I find time.

> My only concern before merging is that you mentioned a `set_o32abi_bit.py` script to fix an issue you had in discord. Is that not needed?

After more investigating, turns out mischief-makers already had a version of this script from ultralib and was partially using it, referenced as `O32_TOOL` in `conifgure.py`. I added an inline comment drawing attention to the fix.

Drahsid:
Oh I do now remember this from back when I had first started matching libultra a couple years ago. Thanks for the clarification, LGTM!

PR #23, opened 2026-05-14
Fix detection, support Linux ARM, cleanup more vars, update README

queueRAM:


configure.py
```diff
@@ -100,7 +91,9 @@ def preferred_tool(programs: List[str]) -> str:
 Z64_PATH = f"build/{BASENAME}.z64"
 OK_PATH = f"build/{BASENAME}.ok"
 
-IDO_DEFS = f"-D_LANGUAGE_C -D_DEBUG -DF3DEX_GBI -DGAME_VERSION={get_version_num()} -DBUILD_VERSION=VERSION_H"
+GAME_VERSION_DEFINE = f"-DGAME_VERSION=GAME_VERSION_{GAME_VERSION.upper()}"
```
queueRAM, line 94: Similar to how `-DBUILD_VERSION=VERSION_H` uses the define instead of number literal, this just uses `-DGAME_VERSION=GAME_VERSION_US1`

configure.py
```diff
@@ -180,25 +174,28 @@ def obtain_ido_recomp(version: str):
         )
         shutil.rmtree(download_dir)
 
-    IDO_RECOMP_VERSION = "v1.1"
+    IDO_RECOMP_VERSION = "v1.2"
```
queueRAM, line 177: Needed to bump to 1.2 to support linux-arm below

configure.py
```diff
@@ -586,5 +582,12 @@ def build(
...
     if args.build:
-        build_command = ["ninja", "-v"] if args.verbose else ["ninja"]
+        build_command = ["ninja"]
+        if args.verbose:
+            build_command += ["-v"]
+            print(f"{CROSS_AS=}")
+            print(f"{CROSS_LD=}")
+            print(f"{CROSS_OBJCOPY=}")
+            print(f"{CPP=}")
+            print(f"{IDO_CC=}")
```
queueRAM, line 592: Found these strings convenient for debugging, so left them in with `--verbose`

README.md
```diff
@@ -7,28 +7,33 @@ A in-progress decompilation of Mischief Makers (or Yuke-Yuke!! Trouble Makers, 
...
-- ninja-build
+- python3 venv
+- ninja
 
+### Linux Dependencies
 Under a Debian based distribution, you can install these with the following commands:
 ```
 sudo apt update
-sudo apt install git build-essential binutils-mips-linux-gnu gcc-mips-linux-gnu python3 python3-venv ninja-build
+sudo apt install git build-essential binutils-mips-linux-gnu python3 python3-venv ninja-build
```
queueRAM, line 23: I removed `gcc-mips-linux-gnu` here since it is no longer needed for building since the only thing using it was `cpp`. If this is needed for something else, I can add it back, but just wanted to keep dependencies to a minimum.

README.md
```diff
@@ -39,30 +44,33 @@ git submodule update --init --recursive
 
 ### Install Python requirements
 Create a project venv and install the pinned Python requirements into it:
-```
+```sh
 python3 -m venv .venv
-.venv/bin/python -m pip install --upgrade pip
-.venv/bin/python -m pip install -r requirements.txt
+source .venv/bin/activate
+python3 -m pip install --upgrade pip
```
queueRAM, line 50: activating the venv puts the .venv/bin in your PATH, so can uses these commands. totally a personal preference

Drahsid:
I don't have an ARM system to test on, were you able to verify that?
I was able to test this works normally on my WSL Ubuntu and native Debian setups.

queueRAM:
> I don't have an ARM system to test on, were you able to verify that?

Yes, I tested this under Arch ARM, Debian aarch64, and Raspian and confirmed macOS still works.

Thanks for confirming Debian and Ubuntu WSL still worked. I also see that there are checks in the code for Windows - is that officially supported and should I test that out?

Drahsid:
iirc I half-assed the support for windows for some reason. If it works (or can work with some small effort) that would be pretty awesome

Edit: come to think of it, it probably is pretty functional since most of the tools are cross platform. Probably just needs a cross toolchain that isn't a nightmare to setup.

queueRAM:
I think I'll save the Windows setup for another time. I gave it an attempt with PowerShell/WinGet alone and this got me through the `configure --split`, but as you mentioned, the toolchain is more difficult. I think it might just be easier to go through mingw64 directly and use its package manager since both some version of `cpp` and the MIPS binutils are needed.
```sh
winget install Git.Git Python.Python.3.14 Ninja-build.Ninja
git clone --recurse-submodules https://github.com/Drahsid/mischief-makers.git
cd mischief-makers
python -m venv .venv
.venv\Scripts\Activate.ps1
python -m pip install -r .\requirements.txt
python .\configure.py --setup
python .\configure.py --split
```

Drahsid:
At that point might as well use WSL haha.
Maybe clang cpp works if you install the non-visual-studio variant, but even then, you need as/ld. Maybe the clang variants work there too (if you compile them for mips) but I've never seen anyone try that for decomp. Maybe worth making an issue to note that it's something worth testing in the future.

In any case, let me know if/when this is ready to merge!

queueRAM:
I'm happy with this PR if you are

PR #29, opened 2026-06-03
Decompile 241E0 and map .rodata

queueRAM:
- Decompile 241E0
- Map 241E0 .rodata
- Use more FixedCoord and clean up data symbols


Drahsid:
Can you run the progress script?

queueRAM:
> Can you run the progress script?

Thanks for the reminder, i added a commit with the updated progress.

PR #30, opened 2026-06-05
Graphics: add sprite index header, document "particle" spawn functions, list some hard-coded palette addresses, id 2 actor fields, misc. doc/cleanup

blackgamma7:
- add header for Identified indexes of sprites.
  - includes macro to get frame of animation for applicable sprites.
  - use indexes where applicable
- document functions that spawn particle actors
- start adding labels to graphic lists
- fix some existing compiler warnings.
- add and apply some hard-coded palette addresses
- use `palette_18C` and `pfn_17C` where applicable and matching.
  - fix casting or data for mis-ID'd palettes.
- id 2 fields in actor struct as timer for i-frames and damage taken in a tick.
- add comments on scripts of overlay code citing which is used where.
- misc. cleanup


include/graphicIndex.h
Drahsid, line 1: I think you should look at the segment -> rle decompression path to understand how sprites are loaded. It's scattered across different rsp segments; so redundant indices are less 'X also uses Y' and more 'X points into a different resource than Y'

include/graphicIndex.h
```diff
@@ -0,0 +1,248 @@
...
+
+#define GINDEX_CLANCERGHOST    0x02A2 // 6 animation frames
+
+#define GINDEX_CATHEAD         0x02CC // 6 animation frames
+#define GINDEX_TEXTBOX         0x02CE // index of the textbox 
+#define GINDEX_SOLIDSQARE      0x02D0 // used as a fill for transisions
+#define GINDEX_ALPHASTART      0x02D2 // Beginning of "Alphabet" charset graphics
+
+
+// indecies of 0x800 and beyond seem subject to change based on scene.
```
blackgamma7: > so redundant indices are less 'X also uses Y' and more 'X points into a different resource than Y'

How can this comment be changed to better reflect this?

include/graphicIndex.h
blackgamma7: I am open on how to reflect this in the header.

include/graphicIndex.h
```diff
@@ -0,0 +1,248 @@
...
+#define GRAPHIC_INDEX_H
+
+
+// indecies used by "graphicIndex" feild of Actor and Portrait structs
+// many graphic indecies use even entries for graphics facing one size,
+// and odd facing the other. for more complex aninmations,
+// Like those of Marina, Clancers, and Beastector, they instead increment normally.
+
+#define GINDEX_LEFT 1 // odd entries are the horizontal mirror of the previous
+#define GRAPHIC_FRAME(g,n) GINDEX_##g##+(2*n) // when we need to declare with animation graphic.
```
Drahsid: If you use a modern preprocessor, this is invalid.

gcc:
```
gcc -E -P -I. /tmp/test.c -o /dev/null
In file included from /tmp/test.c:1:
./include/graphicIndex.h:11:28: error: pasting "GINDEX_BOOM1" and "+" does not give a valid preprocessing token
   11 | #define GRAPHIC_FRAME(g,n) GINDEX_##g##+(2*n) // when we need to declare with animation graphic.
      |                            ^~~~~~~
/tmp/test.c:3:13: note: in expansion of macro ‘GRAPHIC_FRAME’
    3 | int value = GRAPHIC_FRAME(BOOM1, 3);
      |             ^~~~~~~~~~~~~
```

clang:
```
clang -E -P -I. /tmp/test.c -o /dev/null
/tmp/test.c:3:13: error: pasting formed 'GINDEX_BOOM1+', an invalid preprocessing token
    3 | int value = GRAPHIC_FRAME(BOOM1, 3);
      |             ^
./include/graphicIndex.h:11:38: note: expanded from macro 'GRAPHIC_FRAME'
   11 | #define GRAPHIC_FRAME(g,n) GINDEX_##g##+(2*n) // when we need to declare with animation graphic.
      |                                      ^
1 error generated.
```

That said, I'm not sure this macro is better over something explicit like:
```c
#define GINDEX_RIGHT 0
#define GINDEX_LEFT 1
#define GINDEX_DIRS 2

frame = [base] + [left/right] + [index * dirs]
```

src/241E0.c
Drahsid, line 1: D_800BE6C0/D_800BE6CC/D_800BE6D8 (gActorDepthFront/Middle/Back) are still written to here.

include/graphicIndex.h
Drahsid: Maybe something like `GINDEX_SEG[XX]_[RLE FILE]_[NAME]` - see 26A00.c to better understand RLE decompression to the specific addresses that are used for rendering.

versions/us1/symbol_addrs_functions.txt
Drahsid, line 1: Not related to the changes here, but might as well fix it: `Sound_AddSfx = 0x800033020; // type:func` has too many digits

src/overlay_1/overlay_719700/719720.c
```diff
@@ -13,10 +15,10 @@ void func_8019BD70_71A390(s32 arg0) {
 }
 
 void func_8019BD78_71A398(u16 actor_index) {
-    gActors[actor_index].graphicList = D_801A3AEC_72210C;
-    gActors[actor_index].health = 0x12C;
+    gActors[actor_index].graphicLists = D_801A3AEC_72210C;
+    gActors[actor_index].health = 300;
     gActors[actor_index].flags &= 0xFF3CFFFF;\
-    gActors[actor_index].flags |= 0x800000;
+    gActors[actor_index].flags |= ACTOR_FLAG3_UNK23;
```
Drahsid: `ACTOR_FLAG`

include/Alphabet.h
```diff
@@ -431,7 +431,7 @@
 
 #define ALPHA_NULL 0x8FFF
 
-#define ALPHA_GLYPH_INDEX(c) ( (c * 2) + 0x2d2) // get graphic index of ALPHA_* index
+#define ALPHA_GLYPH_INDEX(c) ( (c * 2) + GINDEX_ALPHASTART) // get graphic index of ALPHA_* index
```
Drahsid, line 436: Need to include graphicIndex.h

include/graphicIndex.h
```diff
@@ -0,0 +1,248 @@
...
+#define GINDEX_PORTRAIT_THEOA   0x01FC
+#define GINDEX_PORTRAIT_THEOB   0x01FE
+#define GINDEX_PORTRAIT_THEOC   0x0200
+#define GINDEX_PORTRAIT_CAT     0x0202
+#define GINDEX_PORTRAIT_LUNARA  0x0204
+#define GINDEX_PORTRAIT_LUNARB  0x0206
+#define GINDEX_PORTRAIT_LUNARC  0x0208
+#define GINDEX_PORTRAIT_MERCOA  0x020A
+#define GINDEX_PORTRAIT_MERCOB  0x020C
+#define GINDEX_PORTRAIT_MERCOC  0x020C
```
Drahsid: These are the same

include/common_structs.h
```diff
@@ -15,18 +15,10 @@ typedef struct {
...
-    PORTRAIT_FLAG_UNK1 = (1U << 1U),
-    PORTRAIT_FLAG_PALETTE = (1U << 9U),
-    PORTRAIT_FLAG_UNK14 = (1U << 14U)
-} PortraitFlags;
-
 // used for the portrait transition effect, lifebar, and lifebar head.
 typedef struct {
     /* 0x00 */ Mtx translateMtxs[2];
-    /* 0x80 */ u16 flags; // uses PortraitFlags
+    /* 0x80 */ u16 flags; // uses ActorGFlags
```
Drahsid, line 24: Now there is domain confusion between this and line 29.

include/actor.h
```diff
@@ -166,12 +167,12 @@ typedef struct {
     /* 0x0DE */ u8 unk_0DE;
     /* 0x0DF */ u8 unk_0DF;
     /* 0x0E0 */ s16 health; // initialized from the actor type table and decremented/clamped by damage code
-    /* 0x0E2 */ s16 unk_0E2;
+    /* 0x0E2 */ s16 damageDelta; // damage taken in tick. used in knockback calculation.
```
Drahsid: I think `pendingDamage` might be better since this is also used for knockback

include/data_symbols.h
```diff
@@ -8,7 +8,7 @@ extern u16 D_800D28E4;
 extern u16 D_800D28E8;
 extern u16 D_800D28F0;
 extern u32 D_800D28FC;
-extern u16 D_800D2908;
+extern u16 D_800D2908; // boolean set during "continue" and "attract" states.
```
Drahsid: I believe this skips stage intro and changes the initial load behavior.
It is Set to 1 in GameState_Attract before GameState_Loading, and Set to 1 in continue-screen asm before restarting/loading the stage. It is Cleared only in normal loading when gGameState == GAMESTATE_LOADING. It seems to be read by many stage overlay init/state machines to skip normal opening setup/cutscene paths.

Drahsid (commented):
Great work, sorry for the delay on the full review.

Can you double check that flags, graphicflags, and flags_98 are all using the right enums? I think there may be some preexisting mistakes.

include/common_structs.h
blackgamma7: I could rename the enum `GraphicFlags` and the prefixes `GFXFLAG_*` and move them to `common_enums.h`.

include/graphicIndex.h
blackgamma7: I am concerned this may make them too verbose, what if it was without the `SEG[xx]` or the `[RLE FILE]` parts?

include/graphicIndex.h
blackgamma7: Including the left/right facing seems overkill, since that seems to get set by the `ACTOR_FLAG_FLIPPED` check. I was not aware of the issues with GCC since we have been using the IDO compiler. Is there an `#ifdef` I could use to differentiate between the two?

include/graphicIndex.h
Drahsid: I think just SEG is probably fine, unless it's most common for duplicate indices within the same segment's RLE files (I don't think this is the case)

include/graphicIndex.h
Drahsid: > Including the left/right facing seems overkill, since that seems to get set by the ACTOR_FLAG_FLIPPED check.

At the statement level, is this always right-facing then? In that case I agree that `frame = [base] + [index * dirs]` does make more sense.

You can use `__GNUC__` and `__clang__` and `_MSC_VER` to detect the popular modern compilers/preprocessors; though ideally the macro just works. I think you can probably simplify this to `#define GRAPHIC_FRAME(name, frame) (GINDEX_##name + (2 * (frame)))` to work on all preprocessors, this might be the best option for now.

include/common_structs.h
Drahsid: I'm not familiar enough with the PortraitStruct-relevant code. Is `PORTRAIT_FLAG_PALETTE` == `ACTOR_GFLAG_PALETTE`? If that's the case then you should just need to update the comment on line 29.

blackgamma7: for the bits I observed being set and read in the Portrait draw function (scale, palette, etc.), they seem to match the values and observed effect of the actors' draw function.

So, the comment update should remain?

Drahsid: Yeah I'm not saying this line is wrong. `/* 0x98 */ u16* palette; // used if PORTRAIT_FLAG_PALETTE is set` should probably be `/* 0x98 */ u16* palette; // used if ACTOR_GFLAG_PALETTE is set`, right?

blackgamma7: Ah, Yes! I thought you meant the old line 29.

Drahsid:
From fullclean, there are errors compiling `src/241E0.c` after merge. When this is fixed we can merge.

blackgamma7:
merge fix and a bit more cleanup.

Should be ready.

PR #32, opened 2026-06-13
Text: split script, add data and symbols, propose localization macro.

blackgamma7:
- split code related to displaying game text into separate script.
  - add data related to script.
  - name functions and data related to script.
-  clean up Title screen "Press Start" / copyright code.


include/Alphabet.h
```diff
@@ -435,5 +436,23 @@
...
+#define ALPHA_ARRAY(name,jp,en) u16 ##name [] = en
+/* example:
+  ALPHA_ARRAY(D_800CA280,
+   // "コ-スからでる" / "exit course"
+   {ALPHA_JP_KATA_KO,ALPHA_MINUS,ALPHA_JP_HIRA_KA,ALPHA_JP_HIRA_RA,
+   ALPHA_JP_HIRA_DE,ALPHA_JP_HIRA_RU,ALPHA_NULL},
+   // " exit"
+   {ALPHA_SPACE,ALPHA_EN2_UPPER_E,ALPHA_EN2_LOWER_X,ALPHA_EN2_LOWER_I,ALPHA_EN2_LOWER_T,ALPHA_NULL}
+);*/
+
```
Drahsid: I think this might work better:
```
#if GAME_VERSION == GAME_VERSION_JP
#define ALPHA_PICK(jp, en) jp
#else
#define ALPHA_PICK(jp, en) en
#endif

#define ALPHA_ARRAY(name, jp, en) u16 name[] = ALPHA_PICK(jp, en)
```

I suppose the `u16 name[]` part is also kinda redundant since it doesn't change between versions, so this could be simplified to just ALPHA_PICK, while having the declaration remain more explicit:
```
u16 D_800CA280[] = ALPHA_PICK(
    // "コ-スからでる" / "exit course"
    { ALPHA_JP_KATA_KO, ALPHA_MINUS, ALPHA_JP_HIRA_KA, ALPHA_JP_HIRA_RA, ALPHA_JP_HIRA_DE, ALPHA_JP_HIRA_RU, ALPHA_NULL },
    // " exit"
    { ALPHA_SPACE, ALPHA_EN2_UPPER_E, ALPHA_EN2_LOWER_X, ALPHA_EN2_LOWER_I, ALPHA_EN2_LOWER_T, ALPHA_NULL }
);
```

What are your thoughts?

Drahsid:
I don't see any issues here besides the discourse on the loc macro. If it's alright with you, we can merge this after your other PR.

include/Alphabet.h
blackgamma7: Maybe if `ALPHA_PICK` was called something like `LOCALIZE` so it could be used in other areas of different data. I'd prefer to have the datatype in the macro for the APLHA_* strings just to save on typing it out.

blackgamma7:
Yeah, let's get it in, since we're not needing the macro, yet.

include/Alphabet.h
Drahsid: You're right about naming my suggestion something like `LOCALIZE` instead, it makes more sense in more contexts.

I think the data is going to typically already be there and typed out before rewritten as the localized version, so I'm not sure it actually saves or wastes any time. I personally prefer the explicit declaration, but I can see why you might prefer the other variant. Maybe @queueRAM has some thoughts or opinions on this?

src/text.c
```diff
@@ -0,0 +1,434 @@
...
+       6,5,5,5,5,4,5,5,2,5,5,2,7,5,5,5,5,5,5,5,5,5,6,5,5,5,
+    // A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
+       6,6,6,6,6,6,6,6,3,6,6,5,8,6,6,6,6,6,6,6,6,7,8,6,6,6
+};
+
+//              RGB vals: 0,0,0 7,7,7  15,15,15                         
+u16 gTextPaletteBase[]={0,0x001,0x39cf,0x7bdf};
+
+// set a text character actor.
+// @param actor_index index of acctor.
```
queueRAM: nit: `actor`

include/Alphabet.h
queueRAM: I'm only familiar with what was done in the SM64 decomp which may or may not help here. The project uses custom charsets defined for the dialogs and dialog IDs used throughout. The implementation used UTF-8 strings embedded in the C code with a pre-processing stage with a custom [textconv](https://github.com/n64decomp/sm64/blob/master/tools/textconv.c) tool that replaced anything in `_("")` with raw bytes as defined by the charmap. Each translation had its own [dialogs.h](https://github.com/n64decomp/sm64/blob/master/text/jp/dialogs.h) and [courses.h](https://github.com/n64decomp/sm64/blob/master/text/jp/courses.h) header that defined the strings.

```c
// us/courses.h
SECRET_STAR(COURSE_SA, _("   THE SECRET AQUARIUM"))
// jp/courses.h
SECRET_STAR(COURSE_SA, _("　　　おさかなと　いっしょ"))
```

Drahsid: I've seen other projects do something similar, and this is probably the most readable option. We can turn this into a side-project if @blackgamma7 agrees. I'll probably write a python tool to do the pre-processing.

blackgamma7: I'd wonder how to do this with not only the unique charset, but also its 3 different versions of the English alphabet. Not to mention we'd need to also add the indexes for Kanji, either way.

Are the strings in the SM64 project in the code segments, or a table that's loaded in? My other concern is the ALPHA_* strings are in the scripts themselves.

Either way, if there are no more issues with the PR, we can discuss this elsewhere and get it merged.

Drahsid: Do the different charsets get interleaved in the same strings? If not, that can simplify it.

blackgamma7: I wanna say "No", but I don't want to rule out the possibility. It's possible the original Japanese uses either of the 2 English fonts within the japanese.

Drahsid:
Remove the loc macro for now, we'll tackle it separately. I'll merge after.

blackgamma7:
Done. Ready when you are.

Drahsid:
looks like Alphabet.h is missing an `#endif`

PR #34, opened 2026-06-29
Decompile 59EA0, 5D120, 5E230, 60CF0, and map .rodata

queueRAM:
- Decompile 59EA0, 5D120, 5E230, 60CF0
- Map .rodata for 59EA0, 5E230, 60CF0


src/5E230.c
```diff
@@ -1,109 +1,1060 @@
...
+}
 
-#pragma GLOBAL_ASM("asm/nonmatchings/5E230/func_8005DA5C.s")
+void func_8005D8D0(u16 arg0, u16 arg1) {
+    D_8019B000[4](arg0, arg1);
+}
 
-#pragma GLOBAL_ASM("asm/nonmatchings/5E230/func_8005DA98.s")
+void func_8005D90C(u16 arg0, u16 arg1) {
+    D_80192060[0](arg0, arg1); // TODO: which base? address is D_80192060
```
Drahsid: overlay slot 0 is 0x80192000->0x801920FF for dispatch, so this is 80192000[24]

src/5E230.c
```diff
@@ -1,109 +1,1060 @@
...
-#pragma GLOBAL_ASM("asm/nonmatchings/5E230/func_8005D6DC.s")
+extern s16 D_800D5838[];
+extern s8* D_800D73C0[];
 
-#pragma GLOBAL_ASM("asm/nonmatchings/5E230/func_8005D718.s")
+#define D_8019B000 ((Actor2Func*)0x8019B000)
+#define D_80192020 ((Actor2Func*)0x80192020)
+#define D_80192030 ((Actor2Func*)0x80192030)
+#define D_80192060 ((Actor2Func*)0x80192060)
+#define D_801A6840 ((Actor2Func*)0x801A6840)
```
Drahsid: dispatch table start addrs:
overlay_0: 0x80192000
overlay_1: 0x8019B000
overlay_2: 0x801A6800
overlay_3: 0x801B0800
overlay_4: 0x801B9800

Drahsid (commented):
do this match when using the base address and correcting the indices?

src/5E230.c
queueRAM: thanks! i converted these to use the dispatch table macros and corrected the indexes

src/5D120.c
```diff
@@ -1,39 +1,320 @@
...
+        default:
+            gCamShakeMag -= gCamShakeMagDelta;
+            break;
+        }
+        if (((D_800CCC78[8 * gCurrentScene] == 1) && (gCamShakeV < 0)) || 
+            ((D_800CCC78[8 * gCurrentScene] == 2) && (gCamShakeV > 0))) {
+            gCamShakeV = 0;
+        }
+    }
+}
```
Drahsid, line 135: `gCamShakeMag / 0x10000` looks like fixed point.

queueRAM: To confirm I understand what action to take, should I change `gCamShakeMag` from `s32` → `FixedCoord`? Does anything else need to be done with the `/ 0x10000`?

Drahsid: I'm saying the operation `gCamShakeMag / 0x10000` looks like it is fixedcoord. This may mean the lines looks something like `gCamShakeV = gCamShakeMag.raw / FIXED_UNIT(1.0);`

queueRAM: Got it, thanks for the clarification.

queueRAM: Per discussion, I added `FROM_FIXED(value)` macro to convert from fixed point raw to clarify the intent for the files in this PR.

src/59EA0.c
```diff
@@ -1,60 +1,1008 @@
...
+        gActors[actor_1].graphicFlags |= ACTOR_GFLAG_ROTZ | ACTOR_GFLAG_SCALE;
+        gActors[actor_1].graphicIndex = 0xD0;
+        gActors[actor_1].colorA = 0xFE;
+        gActors[actor_1].var_154 = 4;
+        gActors[actor_1].unk_164 = -4;
+        gActors[actor_1].unk_148 = 8.0f;
+        gActors[actor_1].var_160 = Rand() << 0x12;
+        gActors[actor_1].scaleX = ((f32) func_8005C708(0) * 0.0078125 * 0.3) + 0.5;
+        gActors[actor_1].scaleY = ((f32) func_8005C708(0) * 0.0078125 * 0.1) + 0.2;
+        gActors[actor_1].timer_110 =  (f32) func_8005C708(0) * 0.0078125;
```
Drahsid, line 882: are these casts implicit?

src/5E230.c
Drahsid, line 1: You can probably name a lot of these functions based on context, for example:
```
void func_8005D630(u16 arg0) {
    OVERLAY1_DISPATCH_ACTOR2[3](arg0, 0);
}
```
becomes something like
```
void OverlayABI_Slot1_fn3_v_u16(u16 arg0) {
    OVERLAY1_DISPATCH_ACTOR2[3](arg0, 0);
}
```
The format pretty much being something like `[overlay]_[slot]_[index]_[args]`
If there aren't any redundant uses with incompatible args, those are probably not needed.

src/5E230.c
```diff
@@ -1,109 +1,1057 @@
...
+    OVERLAY0_DISPATCH_ACTOR2[8](arg0, 0);
+}
 
-#pragma GLOBAL_ASM("asm/nonmatchings/5E230/func_8005E09C.s")
+void func_8005DDFC(u16 arg0) {
+    OVERLAY0_DISPATCH_ACTOR2[10](arg0, 0);
+}
 
-#pragma GLOBAL_ASM("asm/nonmatchings/5E230/func_8005E0B0.s")
+void func_8005DE30(void) {
```
Drahsid: 16-byte boundary with completely different type of code. file split?

Drahsid (commented):
looking good, no real issues.

src/59EA0.c
queueRAM: No, `s32 func_8005C708(s32)` and the cast to f32 is needed before implicit cast to `f64` for the multiplication to match. so either the devs tried to be helpful with a cast or there was some f32 variable in between, but i can only get the cast to match: https://decomp.me/scratch/fQdeu

src/5E230.c
queueRAM: Presumably, these will then be renamed with their true meaning when that is understood, right?

How should I handle functions that have the same underlying overlay function? e.g.:
```c
void func_8005D6A0(u16 arg0, u16 arg1) {
    OVERLAY1_DISPATCH_ACTOR2[5](arg0, arg1);
}

void func_8005D780(u16 arg0, u16 arg1) {
    OVERLAY1_DISPATCH_ACTOR2[5](arg0, arg1);
}
```

Drahsid: 💀
For dupes, check if either is unused. If so, you can probably append `unused` or something to the end. Otherwise, you can name them by order I guess, like `_0, _1, _2` etc.

As for renaming with their true meaning, that might end up being tough to disambiguate if the different overlays within each slot vary in intention.

src/5E230.c
queueRAM: Looks like a good idea. I split it at 0x05EA30 and get `OK!`. Testing and adding files now.

src/5E230.c
queueRAM: I renamed these functions. There are a handful of duplicates so for those pairs, I used the `_0`/`_1` suffix. I also updated the progress to reflect these changes. Let me know if you have any more ideas for improvement.

Drahsid:
Incredible. lgtm

PR #35, opened 2026-06-29
Actor: Start list of actor types, add .data to Actor init, document some types, parse more spawn tables.

blackgamma7:
- create list of Actor Types in actorype.h.
  - includes list of overlay id's in high byte to be filled out
- add .data to actor_init.c, with list of cooresponding actor types.
- document behavior of several actor types.
  - Warp gate (aka Warp Star)
  - Clanpot
  - Spikeballs
 - parse actor spawn tables in Overlay 4 scripts.
 - ID or doc more actor fields.
   - 2 are used in clanpot code as inventory check function pointers.
   - one is used to hold warp gate's destination.
   - one is likely the index of the parent actor.
   - 0x110 is another "general purpose" field, set as a param on some init's.
   - 0xD8 is another "general purpose" field, set as a param on some init's.
 - add more labels to actor-related code.
 -  use (u)intptr_t when pointer-as-int is needed to match.
 - add minor documentation for dialouge-related script.
    - add and apply definitions for "text grunt" SFX's.
 - add `#define` for Marina and Teran player actors.
 - cleanup/proofreading.


include/actor.h
```diff
@@ -188,13 +189,22 @@ typedef struct {
...
     /* 0x0F4 */ FixedCoord velocityZ; // applied to posZ in func_80014af0
+
+    // the following values are used by actors for purposes depending on their type
+    // examples: Clanbombs will use offset 0x150 for the fuse timer,
+    // or the crosshair actor will use it for the index of its "parent",
+    // or player-controlled Teran will count his jumps with it.
+    
+    // NOTE: trying to use unions for these variables for different datatypes (u8[4],u16[2])
+    //  may cause mismatch due to "narrowing"
+
```
Drahsid, line 203: I know this is the same comment as below, but moved, but it may be worth better explaining the issue: narrowing is when a wider data type is truncated to fit a smaller data type, ex: s32 -> s16

iirc the discussion on this previously was about a circumstance of 16/8 -> 32 -> 16/8 implied that the source type the compiler saw was smaller, but a literal resulted in int-promotion (ex: lhs is 16 bit but the literal is inherently 32-bit, so it promotes the lhs to 32 bit) which then immediately got narrowed. The tricky part about this is that it caused confusion about the field type because the conversion and odd scheduling around the implicit cast.

include/actorTypes.h
Drahsid, line 1: the spacing around `{}` and `//` and `=` are inconsistent

include/inttypes.h
```diff
@@ -8,4 +8,8 @@
 typedef u32 UNK_TYPE;
 typedef u32 UNK_POINTER;
 
+typedef s32 intptr_t; // sometimes, pointers need to be treated as int's to match.
+typedef u32 uintptr_t; // sometimes, pointers need to be treated as int's to match.
```
Drahsid, line 12:
```suggestion
typedef s32 sptr; // sometimes, pointers need to be treated as int's to match.
typedef u32 uptr; // sometimes, pointers need to be treated as int's to match.
```
I think this reads better and is more consistent with existing primitive type typedefs

include/Scene.h
```diff
@@ -81,7 +81,7 @@
 #define SCENE_CLANCEWAR2        0x4D
 #define SCENE_BEESTHEONE        0x4E
 #define SCENE_INNERSTRUGGLE     0x4F
-#define SCENE_UNK80             0x50 // crash. data suggest more stages were planned for world 5.
+#define SCENE_UNK80             0x50 // crash. data suggest more stages were planned for world 5. Also index of "level clear" graphics.
```
Drahsid: There are 88 scenes. I'm pretty sure this one is a valid scene. Maybe it requires a certain gamestate? It has mostly valid data anyway.

This is what I see in the data at least:
- It's actual collision data is set up coherently
- The functions which are called when the scene is loaded are not null
- It builds the scene collision with collision pages 0000, 0001, 0002 (these blank pages)
- asset rows are populated (besides 4/5, which is unusual I guess)

Since the collision is just pages of zero-bytes, I suspect if you load in with the player, you either get a 'missing 0xC0 blocks' crash or some sort of render crash with bogus assets.

include/actor.h
```diff
@@ -278,26 +286,38 @@ typedef struct {
...
+extern ActorFunc D_80192000[]; // "overlay 0" dispatch table.
+extern ActorFunc D_8019B000[]; // "overlay 1" dispatch table.
+extern ActorFunc D_801A6800[]; // "overlay 2" dispatch table.
+extern ActorFunc D_801B0800[]; // "overlay 3" dispatch table.
+
+
+#define PLAYER_INDEX 0
+
+// actor 0 is always the player character, Marina
+#define gMarina gActors[PLAYER_INDEX]
```
Drahsid:
```suggestion
#define gPlayerActor gActors[PLAYER_INDEX]
```
the PC isn't always Marina. Teran also uses this slot (as well as a couple more,) so 'player actor' is probably the better name here.

Drahsid (commented):
Maybe worth doing a small cleanup pass for whitespace, but otherwise insanely good work!

This one is pretty big, so I probably missed something, @queueRAM care to take a second look?

include/actor.h
blackgamma7: I looked at the memory in that area while playing Teran's levels - the type and HP remained Marina's. I think there's some not-yet documented trickery to hide them and give to control to Teran.

include/inttypes.h
blackgamma7: I looked through the other decomp projects - they also used the intptr_t typedef. I think using would make it more consistent with them.

queueRAM:
> This one is pretty big, so I probably missed something, @queueRAM care to take a second look?

This is beautiful. I skimmed through all your changes and no issues stood out. I agree that `gPlayerActor` makes sense for `gActors[0]`. Note, I think there are some instances of `gActors->` in the codebase that should also be changed, but i've seen this will sometimes no longer match when switching to `gActors[0]`. If you hit this, I'm happy to help look into it.

queueRAM (approved):

blackgamma7:
>  I agree that `gPlayerActor` makes sense for `gActors[0]`.

As noted in a previous comment, `gActors[0]` retains its HP and type during Teran's stages, plus the GemCollect() code suggests it will also change behavior if D_800D2950 is non-zero, suggesting this is the index of the "guest" player character.

> Note, I think there are some instances of `gActors->` in the codebase that should also be changed, but i've seen this will sometimes no longer match when switching to `gActors[0]`. If you hit this, I'm happy to help look into it.

Yeah, I didn't want to touch those just yet, lest I break something.

include/actor.h
blackgamma7: added a clarification in the comment.

blackgamma7:
Added a little documentation of the dialouge script. will expand on a future PR.

Drahsid:
> > I agree that `gPlayerActor` makes sense for `gActors[0]`.
> 
> As noted in a previous comment, `gActors[0]` retains its HP and type during Teran's stages, plus the GemCollect() code suggests it will also change behavior if D_800D2950 is non-zero, suggesting this is the index of the "guest" player character.
> 
> > Note, I think there are some instances of `gActors->` in the codebase that should also be changed, but i've seen this will sometimes no longer match when switching to `gActors[0]`. If you hit this, I'm happy to help look into it.
> 
> Yeah, I didn't want to touch those just yet, lest I break something.

If you modify the position values of actor index 0 on any level, it moves the entire actor the player is in control of. Actor index 0 is likely an invisible root part for Teran, yes, but that doesn't make it not the player actor. I think `gPlayerActor` fits better; or something more verbose like `gPlayerRootActor`

src/28EF0.c
```diff
@@ -4208,97 +4248,139 @@ void func_80034D14(u16 actor_index) {
     gActors[actor_index].scaleY = gActors[actor_index].unk_16C / 1000.0f;
 }
 
-void func_80034D80(u16 actor_index, u16* vals, u16 unused_a2) {
+// load list of items into Clanpot.
+// @param actor_index index of Clanpot.
+// @param vals list to load, ending with 0xFFFF
+// @param unused_a2 unused. causes prototype mismatch.
+void Clanpot_InitItems(u16 actor_index, u16* vals, u16 unused_a2) {
```
Drahsid, line 4250: Does this start at a 16-byte boundary? Maybe clanpots have their own TU?

blackgamma7:
> > > I agree that `gPlayerActor` makes sense for `gActors[0]`.
> > 
> > 
> > As noted in a previous comment, `gActors[0]` retains its HP and type during Teran's stages, plus the GemCollect() code suggests it will also change behavior if D_800D2950 is non-zero, suggesting this is the index of the "guest" player character.
> > > Note, I think there are some instances of `gActors->` in the codebase that should also be changed, but i've seen this will sometimes no longer match when switching to `gActors[0]`. If you hit this, I'm happy to help look into it.
> > 
> > 
> > Yeah, I didn't want to touch those just yet, lest I break something.
> 
> If you modify the position values of actor index 0 on any level, it moves the entire actor the player is in control of. Actor index 0 is likely an invisible root part for Teran, yes, but that doesn't make it not the player actor. I think `gPlayerActor` fits better; or something more verbose like `gPlayerRootActor`

I feel making the the `#define` MORE verbose would defeat the purpose - it's meant to be an abbreviation for commonly accessed data.

Drahsid:
> I feel making the the `#define` MORE verbose would defeat the purpose - it's meant to be an abbreviation for commonly accessed data.

I don't think that's the point at all, considering `gActors[0]` is already quite narrow; if a macro for this exists, it's for readability, since it is not required for matching.

blackgamma7:
In that case, `gMarina` is still more readable:
- This actor's type never changes beyond the usual player character's.
- It is just brief enough to describe which actor type it always remains as.
- `gPlayerActor` can be confusing as there are instances where Marina is not the player character, and the current one exists in another index.
- the "root" in  `gPlayerRootActor` seems redundant, and I believe the length is counterintuitive to a purpose of readability.
- `gLifebar` and `gLifebarHead` have been similarly brief `#define`'s that were adequately readable.
- `gMarinaActor` would be another possibility.

Regardless, I still need to do a merge fix.

Drahsid:
Based on my understanding, `gActors[0]` directly drives `gPlayerPosX/Y`, `gPlayerVelX/YMirror` all gameplay camera-relative positioning, all player collision checks (including guests!), and the default state where there isn't a guest character. Importantly, Teran's other actor(s), for example, derive their positions from `gActors[0]`, it gets directly copied, then parts are attached at relative points.

This is why I think `gPlayerRootActor` is the most accurate name. I agree it's verbose, but that's why I think `gPlayerActor` is a good compromise. I would argue that the code makes it clear that `gMarina` is more wrong:
- If `gMarina` is chosen, then all player derived globals have incoherent names (ex: `gPlayerPosX/Y = gMarina.posX/Y + screenX/Y;` read as being wrong at the assignment. If they are changed to match the 'Marina' wording, they would then not actually explain what is happening; where the neutral `player` syntax does.
- In Teran mode, slot 0 doesn't behave as Marina, which is a contradiction to the suggested name. If you observe `gActors[0]` in this state, she would be frozen in the grabbing state; but in spite of that, `gActors[0]` still directly drives the collision and physics.
- "This is Marina" does describe what `gActors[0]` typically contains, but "This is the player's root/primary actor" describes the actual surrounding behavior in all instances.

I think the real win here is to name `D_800D2950` something relevant like `player's guest actor index`, as that solves all of the ambiguity and becomes more self-documenting. For example, consider the following prose:
```c
void func_8002F93C(u16 actor_index, u16 is_static, void* palette, s16 pos_x, s16 pos_y) {
    // ...
    // grab blue gem
    else if (palette == D_800D86D8) {
        if (gGuestPlayerActorIndex != 0) {
            gActors[gGuestPlayerActorIndex].health += 30;
        }
        else {
            Actor_IncrementHealth(PLAYER_INDEX, 30); // func_80057320
        }
        // ...
    }
    // ...
}

// func_80057320
s32 Actor_IncrementHealth(u16 actor_index, s16 health_increment) {
    if (gActors[actor_index].health < 0) {
        return -1;
    }

    gActors[actor_index].health += health_increment;
    if (gActors[actor_index].health > 3000) {
        gActors[actor_index].health = 3000;
        return 2;
    }
    else {
        return 1;
    }
}
```

There is no ambiguity here between the roles of the player actor and the guest actor, and that would probably be even more clear in instances with functions that read like:
```c
void HealPlayerActor(s16 amount) {
    if (gGuestPlayerActorIndex) {
        gActors[gGuestPlayerActorIndex].health += amount;
    }
    else {
        gPlayerActor.health += amount;
    }
}

void HealActor(u16 actor_index, s16 amount) {
    if (actor_index == PLAYER_INDEX) {
        if (gGuestPlayerActorIndex) {
            gActors[gGuestPlayerActorIndex].health += amount;
        }
        else {
            gPlayerActor.health += amount;
        }
    }
}
```

You could also probably make it a bit more readable by adding an additional define:
`#define gGuestPlayerActor gActors[gGuestPlayerActorIndex]`

blackgamma7:
I will compromise with `gPlayerActor`.
I also added `#define`'s for Teran's player actor indexes.

include/actor.h
```diff
@@ -247,8 +250,14 @@ typedef struct {
...
-    /* 0x168 */ s32 unk_168;
-    /* 0x16C */ s32 unk_16C;
+    union{
+        /* 0x168 */ s32 unk_168;
+        /* 0x17C */ void(*clanpot_pfn0)(u16 item_type,u16 actor_index); // used by clanpots when mixing to tally items
+    };
+    union{
+        /* 0x16C */ s32 unk_16C;
+        /* 0x17C */ s32(*clanpot_pfn1)(u16 actor_index); // used by clanpots when mixing. returns true if requirements mat.
+    };
```
Drahsid, line 262:
```suggestion
    union {
        /* 0x168 */ s32 unk_168;
        /* 0x168 */ void(*clanpot_pfn0)(u16 item_type, u16 actor_index); // used by clanpots when mixing to tally items
    };
    union {
        /* 0x16C */ s32 unk_16C;
        /* 0x16C */ s32(*clanpot_pfn1)(u16 actor_index); // used by clanpots when mixing. returns true if requirements mat.
    };
```
spacing, and wrong offsets in the comments.
Also, we should typedef the function types.

include/actorTypes.h
```diff
@@ -0,0 +1,443 @@
...
+    ACTOR_OVL3_39,
+    ACTOR_OVL3_40,
+    ACTOR_OVL3_WORLD5B,
+    ACTOR_OVL3_42,
+    ACTOR_OVL0_INTRO, // actors in Overlay 0 code of intro cinematic.
+    ACTOR_OVL0_ENDING // actors in Overlay 0 code of Ending.
+} ActorOverlayIDs;
+
+
+typedef enum{
```
Drahsid:
```suggestion
typedef enum {
```
spacing

include/thread.h
```diff
@@ -15,7 +15,7 @@ extern OSThread gRmonThread;
 
 
 // an error on the developer's part: yeild data is WAY TOO BIG.
-// OS_YEILD_DATA_SIZE (0x900) should be divided by sizeof(u64),
+// OS_YIELD_DATA_SIZE (0x900) should be divided by sizeof(u64),
 // but instead it's made 8 times larger than it should be,
 // taking up a significant ammount of the .bss
 extern u64 gYeildData[0x900];
```
Drahsid: Good catch on the typo. It exists in the variable `gYeildData` too.

src/overlay_0/overlay_67DBA0/67DC20.c
```diff
@@ -235,23 +235,23 @@ void func_801935A4_67F0C4(void) {
...
+    Actor_SetHitboxBX(actor_index, -4, 0x1A);
     gActors[actor_index].hitboxAY0 = 0xC;
     gActors[actor_index].hitboxAY1 = -0xC;
-    func_8002AC7C(actor_index, -0x10, 6);
+    Actor_SetHitboxAX(actor_index, -0x10, 6);
 }
 
 void func_80193728_67F248(u16 actor_index) {
-    ACTOR_INIT(actor_index,0x1D);
+    ACTOR_INIT(actor_index,ACTORTYPE_GRAPHICONLY);
```
Drahsid:
```suggestion
    ACTOR_INIT(actor_index, ACTORTYPE_GRAPHICONLY);
```
spacing.

src/overlay_0/overlay_67DBA0/67DC20.c
```diff
@@ -513,7 +513,7 @@ void func_801949B8_6804D8(s32 arg0) {
 }
 
 void func_80194A38_680558(void) {
-    ACTOR_INIT(0x60,0x1D);
+    ACTOR_INIT(0x60,ACTORTYPE_GRAPHICONLY);
```
Drahsid:
```suggestion
    ACTOR_INIT(0x60, ACTORTYPE_GRAPHICONLY);
```
spacing

src/overlay_0/overlay_6BCD80/6BCDD0.c
```diff
@@ -931,7 +933,7 @@ void func_8019DAF0_6C87C0(u16 arg0, u16 arg1, u16 arg2, u16 arg3) {
 }
 
 void func_8019DBEC_6C88BC(u16 actor_index) {
-    ACTOR_INIT(actor_index,0x1D);
+    ACTOR_INIT(actor_index,ACTORTYPE_GRAPHICONLY);
```
Drahsid:
```suggestion
    ACTOR_INIT(actor_index, ACTORTYPE_GRAPHICONLY);
```
spacing

src/overlay_0/overlay_6BCD80/6BCDD0.c
```diff
@@ -962,7 +964,7 @@ void func_8019DCB8_6C8988(void) {
 #pragma GLOBAL_ASM("asm/nonmatchings/overlay_0/overlay_6BCD80/6BCDD0/func_8019DCE8_6C89B8.s")
 
 void func_8019DF44_6C8C14(u16 actor_index) {
-    ACTOR_INIT(actor_index,0x1D);
+    ACTOR_INIT(actor_index,ACTORTYPE_GRAPHICONLY);
```
Drahsid:
```suggestion
    ACTOR_INIT(actor_index, ACTORTYPE_GRAPHICONLY);
```
spacing

src/overlay_0/overlay_6D1BF0/6D1C00.c
```diff
@@ -95,8 +95,8 @@ void ClancerCommanderMech_Update(u16 actor_index) {
                 gActors[actor_index].velocityX.raw = FIXED_UNIT(4.5);
                 gActors[actor_index].velocityY.raw = FIXED_UNIT(0.75);
                 func_80081790(actor_index, D_800E58D4);
-                ACTOR_INIT(temp_index_22,0x1D);
-                gActors[temp_index_22].var_150 = (s32)D_800E1180;
+                ACTOR_INIT(temp_index_22,ACTORTYPE_GRAPHICONLY);
```
Drahsid:
```suggestion
                ACTOR_INIT(temp_index_22, ACTORTYPE_GRAPHICONLY);
```
spacing

src/overlay_2/overlay_76F390/76F3A0.c
```diff
@@ -92,16 +92,16 @@ void func_801A6B28_76F5C8(u16 actor_index) {
     FlameArchInit* temp_v1;
 
     if ((gActiveFrames & 3) == 0) {
-        func_8005CA34(4, 5);
-        temp_index = gActors[actor_index].unk_0D8 * 5;
+        CameraShake(4, 5);
+        temp_index = gActors[actor_index].var_0D8 * 5;
         temp_v1 = (FlameArchInit*)&((s16*)D_801A7250_76FCF0)[temp_index];
-        temp_v0 = SpawnParticle_Image_90C0_16(0x1B4, gActors[actor_index].posX.whole, temp_v1->posY, temp_v1->posZ + 1);
+        temp_v0 = SpawnParticle_Image_90C0_16(GRAPHIC_FRAME(DASHEFFECT,1), gActors[actor_index].posX.whole, temp_v1->posY, temp_v1->posZ + 1);
```
Drahsid:
```suggestion
        temp_v0 = SpawnParticle_Image_90C0_16(GRAPHIC_FRAME(DASHEFFECT, 1), gActors[actor_index].posX.whole, temp_v1->posY, temp_v1->posZ + 1);
```
spacing

src/overlay_2/overlay_76F390/76F3A0.c
```diff
@@ -113,7 +113,7 @@ void func_801A6CAC_76F74C(u16 actor_index) {
     u16 temp_v0;
 
     if ((gActiveFrames % 5) == 0) {
-        temp_v0 = SpawnParticle_Image_90C0_16(0x1B4, 0, 0, 0);
+        temp_v0 = SpawnParticle_Image_90C0_16(GRAPHIC_FRAME(DASHEFFECT,1), 0, 0, 0);
```
Drahsid:
```suggestion
        temp_v0 = SpawnParticle_Image_90C0_16(GRAPHIC_FRAME(DASHEFFECT, 1), 0, 0, 0);
```
spacing

include/actor.h
```diff
@@ -283,26 +292,46 @@ typedef struct {
     union {
         /* 0x190 */ s32 unk_190;
         /* 0x190 */ void* unk_190_p;
+        /* 0x190 */ u16* warpGate_coords; // coordinates for a warp gate. {x,y,x-facing}
```
Drahsid:
```suggestion
        /* 0x190 */ u16* warpgateCoords; // coordinates for a warp gate. {x,y,x-facing}
```
`camelCase` for known struct fields

src/overlay_4/overlay_7D40C0/7D40F0.c
```diff
@@ -107,76 +107,102 @@ u32 D_801BC178_7D6968[] = {
...
+//  flags   index   posX    posY    0x110   0xD8    type
+    0x0000, 0x008F, 0x0EB8, 0x0170, 0x0081, 0x0000, ACTORTYPE_WARPGATE,
+    SPAWNRECORD_END
 };
 
+// contents of second Clanpot in "Go Marzen 64"
 u16 D_801BC7BC_7D6FAC[] = {
-    0x8400, 0x1111, 0x0000, 0x1C07, 0x0069, 0x8400, 0x1111, 0x0000,
-    0x1C07, 0x0069, 0x8400, 0x1111, 0x0000, 0x1C07, 0x0069, 0xFFFF,
+//  flags                                   0x110   0xD8 type                         Icon
```
Drahsid:
```suggestion
//  flags                                   0x110  0xD8  type                         Icon    
```
seems to be off by 1 here

src/overlay_4/overlay_7E6A20/7E6A50.c
```diff
@@ -65,82 +65,109 @@ u16 D_801BA56C_7E76BC[] = {
...
 
-u32 D_801BA63C_7E778C[] = {
-    0x1000027F, 0x00007FFF, 0x0574020F, 0xFFE0FFEB,
-    0x00000048, 0x100001B0, 0x00000000, 0x1F01FF00,
+u16 D_801BA63C_7E778C[] = {
+    0x1000, 0x027F, 0x0000, 0x7FFF, 0x0574, 0x020F, 0xFFE0, 0xFFEB
+};
+
+// unused spawn table
+u16 D_801BA64C_7E779C[] ={
```
Drahsid:
```suggestion
u16 D_801BA64C_7E779C[] = {
```
spacing

src/overlay_4/overlay_7E6A20/7E6A50.c
```diff
@@ -65,82 +65,109 @@ u16 D_801BA56C_7E76BC[] = {
...
+    { 0x0600, { 0 }, D_801BA704_7E7854, D_801BA714_7E7864, D_801BA700_7E7850, 0x0200, 0x0800 },
+    { 0x1100, { 0 }, D_801BA724_7E7874, D_801BA734_7E7884, D_801BA700_7E7850, 0x0800, 0x1300 },
+    { 0x1E00, { 0 }, D_801BA744_7E7894, D_801BA754_7E78A4, D_801BA700_7E7850, 0x1300, 0x2000 },
+    { 0x2C00, { 0 }, D_801BA764_7E78B4, D_801BA774_7E78C4, D_801BA700_7E7850, 0x2000, 0x2E00 },
 };
 
-u32 D_801BA700_7E7850[] = {
-    0x00000000,
+u16 D_801BA700_7E7850[] = {
+    0,0
```
Drahsid:
```suggestion
    0, 0
```
spacing

include/actorTypes.h
```diff
@@ -0,0 +1,443 @@
...
+    ACTORTYPE_OVL2_W3_21,
+    ACTORTYPE_OVL2_W3_22,
+    ACTORTYPE_OVL2_W3_23,
+    ACTORTYPE_OVL2_W3_24,
+
+    // actors in dispatch table D_801A6800_78B430
+
+    ACTORTPYE_OVL2_W5B_0 = (ACTOR_OVL2_STAGE_5_5 << 8), // stubbed actor
+    ACTORTPYE_OVL2_W5B_PIRATE, // pirate at end of "Trapped"
+    ACTORTPYE_OVL2_W5B_SPIKEBALL, // spikeballs of pirate at end of "trapped"
```
Drahsid:
```suggestion
    ACTORTYPE_OVL2_W5B_0 = (ACTOR_OVL2_STAGE_5_5 << 8), // stubbed actor
    ACTORTYPE_OVL2_W5B_PIRATE, // pirate at end of "Trapped"
    ACTORTYPE_OVL2_W5B_SPIKEBALL, // spikeballs of pirate at end of "trapped"
```
typo

include/actorTypes.h
```diff
@@ -0,0 +1,443 @@
...
+    ACTORTYPE_OVL0_GEN_BOOMERANG,
+    ACTORTYPE_OVL0_GEN_FLAME,
+    ACTORTYPE_OVL0_GEN_BLOCK, // destructible block
+    ACTORTYPE_OVL0_GEN_6, // statue?
+    ACTORTYPE_OVL0_GEN_7,
+    ACTORTYPE_OVL0_GEN_8,
+    ACTORTYPE_OVL0_GEN_9,
+    ACTORTYPE_OVL0_GEN_10,
+
+    // actors in dispatich table D_801B0800_7C0AB0
```
Drahsid:
```suggestion
    // actors in dispatch table D_801B0800_7C0AB0
```
typo

include/actorTypes.h
```diff
@@ -0,0 +1,443 @@
...
+    ACTORTYPE_OVL3_W5_19,
+    ACTORTYPE_OVL3_W5_20,
+    ACTORTYPE_OVL3_W5_TURRET, // turrets in third section of "Trapped"
+    ACTORTYPE_OVL3_W5_22,
+    ACTORTYPE_OVL3_W5_23,
+    ACTORTYPE_OVL3_W5_24,
+    ACTORTYPE_OVL3_W5_25,
+    ACTORTYPE_OVL3_W5_26, // graphic-only actors?
+
+    // actors in dispatch rable D_80192000_6F9330
```
Drahsid:
```suggestion
    // actors in dispatch table D_80192000_6F9330
```
typo

Drahsid (approved):
Great stuff. Pretty much only typos and nitpicks. I think the progress report is out of date, can you verify it's updated? After that, let me know when you're ready for merge.

src/28EF0.c
blackgamma7: the code, yes, the data, no. Checked several versions.

src/overlay_0/overlay_67DBA0/dispatch_table.c
Drahsid, line 1: Maybe we convert dispatch tables to `uintptr_t` to remove all ambiguity. Can probably be in a different commit/PR though.

Drahsid (approved):

Drahsid:
Looks good. Will merge if you're ready.

blackgamma7:
Made a couple last-minutes changes, otherwise ready when you are.

PR #37, opened 2026-07-02
Decompile 5 functions in marina.c

ThiagoLira:
Decompiles five functions in `src/marina.c`:

- `func_80048C28`
- `func_80048C94`
- `func_8004A8E0`
- `func_8004A918`
- `func_8004A960` (Marina init: zeroes `D_801373E0`, resets velocities/scales/flags)

Notes:

- All five functions match (checksum OK)
- Adds `extern` declarations in `marina.c` for `D_801373D8`, `D_800D5794`, and `D_800D54EC` (no changes to symbol address files were needed)
- In `func_8004A960`, the zeroing loop's end bound compiles to the reloc `D_801373E0 + 0x80` rather than the auto symbol `D_80137460` (same address — one past the end of the struct), so the object-level diff shows a reloc name difference but the linked ROM is identical

Disclosure: these were decompiled with AI assistance (Claude Fable), with every function verified byte-for-byte against the US 1.1 ROM.
I'd be ok with changing this PR to conform with any other guidelines, this game is very dear to me.


src/marina.c
```diff
@@ -98,11 +125,43 @@ s32 func_80048CE4(void) {
...
+
+    D_801373E0.unk_13 = 0;
+    D_801373E0.unk_12 = 100;
+    actor->unk_140_u16[1] = 0;
+    actor->unk_140_u8[1] = 0;
+    actor->unk_180 = 0;
+    actor->unk_12F_u8 = 0;
+    actor->graphicLists = D_800D54EC;
+    actor->velocityX.raw = actor->velocityY.raw = actor->velocityZ.raw = 0;
+    actor->graphicFlags |= 0x501;
```
Bl00D4NGEL:
```suggestion
    actor->graphicFlags |= ACTOR_GFLAG_UNK10 | ACTOR_GFLAG_UNK8 | ACTOR_GFLAG_SCALE;
```

src/marina.c
```diff
@@ -98,11 +125,43 @@ s32 func_80048CE4(void) {
...
+    Actor* actor = &gActors[actor_index];
+    s32 i;
 
-#pragma GLOBAL_ASM("asm/nonmatchings/marina/func_8004A918.s")
+    actor->unk_12E_u8 |= 0x41;
 
-#pragma GLOBAL_ASM("asm/nonmatchings/marina/func_8004A960.s")
+    for (i = 0; i < 0x20; i++) {
+        ((u32*)&D_801373E0)[i] = 0;
+    }
```
Drahsid: [Style Guide](https://github.com/Drahsid/mischief-makers/wiki/Style-Guide): prefer wide variable names (ex: 'index' over 'i')

Also, `((u32*)&D_801373E0)[i] = 0;` is a fakematch. There should be no reasons to do pointer hacks here.

src/marina.c
```diff
@@ -98,11 +125,43 @@ s32 func_80048CE4(void) {
...
+    actor->velocityX.raw = actor->velocityY.raw = actor->velocityZ.raw = 0;
+    actor->graphicFlags |= ACTOR_GFLAG_UNK10 | ACTOR_GFLAG_UNK8 | ACTOR_GFLAG_SCALE;
+    actor->flags &= ACTOR_FLAG_FLIPPED;
+    actor->flags |= ACTOR_FLAG_UNK27 | ACTOR_FLAG_UNK16 | ACTOR_FLAG_UNK12 | ACTOR_FLAG_UNK8 | ACTOR_FLAG_ENABLED;
+    actor->unk_0DE = 1;
+    actor->unk_0DF = 1;
+    actor->unk_17C = 0;
+    gActors[0].unk_120 = actor->unk_124 = actor->unk_128 = 1.0f;
+    actor->unk_170 = 1;
+    actor->state = 3;
```
Drahsid: Does `gActors[actor_index].` match over `actor->` ?
In cases where it makes a difference, probably about 90% of the time the right form is array indexing, so we typically choose that unless the pointer variant is required for matching.

Drahsid (commented):
Is this mostly an AI PR? Did you review the changes it made?

src/marina.c
ThiagoLira: Change the variable name to 'index', I also removed the cast there and added a view on the struct like others structs on `common_structs.h`. Does this work better?

src/marina.c
ThiagoLira: It does match! Committed the change

ThiagoLira:
I did review what Fable wrote, I'm a bit rusty with assembly and C. Also learning the guidelines here, thank you for the review!

include/common_structs.h
```diff
@@ -47,7 +47,10 @@ typedef struct {
 } UnkStruct_801069E0; /* sizeof = 0x90 */
 
 // struct storing data about Marina player actor
-typedef struct {
+// cleared word-by-word on Marina init (func_8004A960), hence the raw word view.
+typedef union {
+    u32 raw[0x20];
+    struct {
```
Drahsid: Doubtful this reflects the original source, but I realized my original suggestion was off-base. See below.

src/marina.c
```diff
@@ -98,11 +125,42 @@ s32 func_80048CE4(void) {
...
+void func_8004A960(u16 actor_index) {
+    s32 index;
 
-#pragma GLOBAL_ASM("asm/nonmatchings/marina/func_8004A918.s")
+    gActors[actor_index].unk_12E_u8 |= 0x41;
 
-#pragma GLOBAL_ASM("asm/nonmatchings/marina/func_8004A960.s")
+    for (index = 0; index < 0x20; index++) {
+        D_801373E0.raw[index] = 0;
+    }
```
Drahsid:
```suggestion
    for (index = 0; index < sizeof(D_801373E0) / sizeof(u32); index++) {
        ((u32*)&D_801373E0)[index] = 0;
    }
```

typical memset/bzero pattern seems more likely than a struct being declared as a union for the mere purpose of having data as a byte array. I highly doubt the original programmer decided to count the size of the array manually.

I also apologize for my original suggestion that this is a fakematch, I didn't actually see the bigger picture until I saw the struct it was tied with; so the real context to see over the original variant you had is `sizeof(D_801373E0) / sizeof(u32) == 0x20`

ThiagoLira: No worries, thank you!

Drahsid:
After some deliberation, I've decided to reject this PR.

First and foremost, I want to thank you for your time and want to contribute.

I'm not particularly for or against AI, nor are any of the existing contributors (as far as we have discussed,) but I think there is important context to the reason for my rejection:
- I want to avoid encouraging people to use AI to match trivial functions, as this will quickly leave potential new contributors (who don't rely on AI) with no functions that they can learn from. I don't mind tackling low-hanging fruit, but I think it is in the better hands of any dedicated contributor.
- Having Claude or Codex on the contributor/co-author list comes with a stigma which may attract more vibe-coders and deter people who would get the impression that the 'use of AI = lazy'

This essentially boils down to: I don't think it's worth accepting an AI co-authored PR that only contains a couple small functions.
Note that I would probably be happy to accept an AI co-authored PR which contained a [couple] highly difficult or complex function, where you could both show your work, and how the AI helped you solve confounding issues.
I'm certainly not saying I will outright reject any future contributions from you if you choose to come back with something like that, or manually authored work.

Thanks again!

PR #38, opened 2026-07-03
Decompile 156F0, 1A120, 61B80, 82F80, stage

queueRAM:
- Decompile 156F0, 1A120, 61B80, 82F80, debug_level_select → stage
- Merge part of 17A70 and all of 1A120 into stage in order to map .rodata
- Map all of stage .data
- Map all of 82F80 .bss


src/stage.c
blackgamma7, line 1: Since the data section contains the near-entirety of stage data (time targets, starting music, names, etc.) and not just debug data and code, I'd suggest changing the name of the script and header to reflect this, or even sectioning off the data into its own script.

blackgamma7 (commented):
Something I wanted to change about the "debug_level_select" data, since you are working on it:

src/stage.c
queueRAM: Agree a name name change makes sense given the increase in scope. Do you have any recommendations? Something like `level_select.c` or `levels.c`?

blackgamma7: something  like `stages.c` and `stages.h`, yeah. I'd also like to do a pass of adding the `ALPHA_*` indexes for the name "strings" and add labels, I can do that myself in another PR.

include/rle.h
```diff
@@ -0,0 +1,34 @@
...
+#define ASSET_DEST2           ((u8*)0x80300000)
+#define ASSET_DEST3           ((u8*)0x80320000)
+#define ASSET_DEST4           ((u8*)0x80340000)
+#define ASSET_DEST5           ((u8*)0x80360000)
+#define ASSET_DEST5_TABLE     ((u8*)0x8036BAC0)
+#define ASSET_DEST6           ((u8*)0x803CE800)
+#define ASSET_DEST6_TABLE     ((u8*)0x803D7748)
+#define D_800D0E84_PU32       ((u32*)D_800D0E84) // TODO
+
+extern s32 Trouble_RLE_Type1(u8* src, u8* dst);
```
Drahsid: What about Trouble_RLE_Type2?

include/rle.h
Drahsid, line 1: Maybe worth merging the segment headers into this since they deal with rle assets

src/26A00.c
```diff
@@ -72,30 +67,16 @@ extern u32 D_80137788;
...
+#define OVERLAY0_DISPATCH_TABLE_VRAM (0x80192000)
+#define OVERLAY0_OBJECT_VRAM         (0x80192100)
+#define OVERLAY1_DISPATCH_TABLE_VRAM (0x8019B000)
+#define OVERLAY1_OBJECT_VRAM         (0x8019B100)
+#define OVERLAY2_DISPATCH_TABLE_VRAM (0x801A6800)
+#define OVERLAY2_OBJECT_VRAM         (0x801A6900)
+#define OVERLAY3_DISPATCH_TABLE_VRAM (0x801B0800)
+#define OVERLAY3_OBJECT_VRAM         (0x801B0900)
+#define OVERLAY4_DISPATCH_TABLE_VRAM (0x801B9800)
+#define OVERLAY4_OBJECT_VRAM         (0x801B9900)
```
Drahsid: maybe redundant and worth putting in a header

src/61B80.c
```diff
@@ -1,103 +1,1576 @@
...
+        case 7:
+            gActors[actor_index].colorR = 0x7F; \
+            gActors[actor_index].colorB = 0x7F; \
+            gActors[actor_index].colorG = 0; \
+            break; \
+        case 4:
+        case 8:
+            gActors[actor_index].colorG = 0x7F; \
+            gActors[actor_index].colorR = 0; \
+            gActors[actor_index].colorB = 0; \
```
Drahsid, line 418: think there was mention of this in the past, but, RGB macro?

src/stage.c
Drahsid: 'scene' and 'map' are other names which work

src/debug_level_select.c
```diff
@@ -92,160 +119,2242 @@ u16 gStageTimesToBeat[] = {
...
+    else {
+        var_v0 = 1;
+    }
+    func_80083810(1, 4, D_800C9540, var_v0);
+    func_8001A7E0(6, 4, gStageTime, gCurrentStage, var_v0);
+    func_8001A838(9, 4, gStageTime, gCurrentStage, var_v0);
+}
+#else
+#pragma GLOBAL_ASM("asm/nonmatchings/debug_level_select/func_8001A890.s")
+#endif
```
Drahsid: is there an issue using these prototypes in this TU?

src/stage.c
blackgamma7: I'd caution against 'scene', I picked the name originally as a way to differentiate between data loaded for just stages and everywhere else, like "demo" and intro cinematics.

include/rle.h
queueRAM: Agreed, `Trouble_RLE_Type2` should be added here, but i'm not sure what the interface is since it nor its callers have been decompiled yet. I'll leave it out for now unless you think it has the same interface as 1.

src/debug_level_select.c
queueRAM: I'll take a closer look again, but func_8001A7E0/func_8001A838 are called from multiple functions in this TU and i originally couldn't find a common interface that both match. I also couldn't get the functions themselves to match changing the interface. `func_8001A838` is the function immediately before `func_8001A890` here so it would be odd for them to use different interfaces. Unless there is a split at func_8001A890?

include/rle.h
Drahsid: iirc `void Trouble_RLE_Type2(u8* src, u8* dst, s32 length);`

src/debug_level_select.c
Drahsid: If the same prototype doesn't work, it could indicate a TU split somewhere, or a misplaced definition mistake like:
```
void func1() {
}

void func2() {
    somefunc(1, 2); // prototype bug
}

void somefunc(s32 arg1, s32 arg2);
void func3() {
    somefunc(1, 2);
}
```

queueRAM: This one is puzzling because the functions in question are immediately after each other and call each other. It would be an odd place to have a mismatch and an odd place to split. I'm hoping that it's just difficult to match. Anyway, you've convinced me to match the dependencies to reveal what the underlying interfaces are, so I'm going through 82F80 now.
```c
void func_8001A7E0(s16 arg0, u16 arg1, u16 arg2, u16 arg3, s16 arg4) {
...
}

void func_8001A838(s16 arg0, u16 arg1, u16 arg2, u16 arg3, s16 arg4) {
...
}

void func_8001A890(void) {
    u16 var_v0;
    ...
    func_8001A7E0(6, 4, gStageTime, gCurrentStage, var_v0);
    func_8001A838(9, 4, gStageTime, gCurrentStage, var_v0);
}
```

src/stage.c
queueRAM: I'm happy to rename this with whatever you agree on. I'm not familiar enough with the usage to help form an opinion. If need some time to think, can leave as-is and rename later.

Drahsid: `stage` is probably the best option

src/debug_level_select.c
queueRAM: After matching some of 82F80 and using the prototypes and changing func_8001A7E0/func_8001A838, get `OK!` now. Not sure what was causing my issues before, but resolved it now.

src/stage.c
queueRAM: Renamed `debug_level_select` → `stage`

src/61B80.c
queueRAM: I tried to add the following macro, but sadly it doesn't work here since I think everything needs to be on the same line and sometimes `B` or `G` are assigned first. I think I'll omit this for now.
```c
#define ACTOR_SET_COLOR(index, red, green, blue) \
 gActors[actor_index].colorR = (red); \
 gActors[actor_index].colorG = (green); \
 gActors[actor_index].colorB = (blue);
```

Alternatively, I think i could format the code like this if you prefer:
```c
switch (gActors[actor_index].unk_0D8 & 0xF) {
case 1: case 5: gActors[index].colorB = 0x7F; gActors[index].colorR = 0; gActors[index].colorG = 0; break;
case 2: case 6: gActors[index].colorR = 0x7F; gActors[index].colorG = 0; gActors[index].colorB = 0; break;
...
```

include/rle.h
queueRAM: Referring to the Segment03.h and Segment04.h or something else? I don't think I know these well enough and they include the generated asset headers which may be more involved so I think I'll defer that to a future PR.

src/61B80.c
Drahsid: Your first macro takes 'index' but does nothing with it.

I was thinking the ordering could have been an ido quirk due to scheduling, but if the macro doesn't work then the actual code probably used the one-liner format.

include/common_structs.h
```diff
@@ -6,7 +6,7 @@ typedef union {
     /* 0x00 */ s32 raw;
     struct {
     /* 0x00 */ s16 whole;
-    /* 0x02 */ u16 frac;
+    /* 0x02 */ s16 frac;
```
Drahsid, line 9: did this break anything?

src/61B80.c
queueRAM: > Your first macro takes 'index' but does nothing with it.

Oops, thanks for calling that out, copy-paste error. I tied to reduce the width for the GH message but made them incorrect in both macro and C snippet.

For now, I'm leaving the `\` on multiple lines since it gets really wide as a one-liner, but let met know if you have a preference.

include/common_structs.h
queueRAM: Nothing breaks, still get `OK!`. I think this is the first code that reads from `frac` and it was needed. See `gActors[arg0].posY.frac = gActors[var_a1].posY.frac;` accesses in https://decomp.me/scratch/s22AN

Drahsid: Hmm. This does seem to imply that, but that creates some confusion about the coordinate space...

Edit: looking at more disassembly around this issue, it seems to be a genuine mistake from the original programmer(s). They don't use this often, and its typically for copying, zeroing, or resetting; so I suppose they just never noticed.

Drahsid:
Looking good, ready to merge?

queueRAM:
> Looking good, ready to merge?

Yes, ready to merge at your convenience.

PR #39, opened 2026-07-06
Decompile 12DD0, actor_init, gameover, lifebar, soft_reset

queueRAM:


include/common.h
```diff
@@ -37,7 +37,7 @@ typedef s32 DEFAULT_INT; // use this to explicitly show when the original progra
 #define SCREEN_WIDTH (320)
 #define SCREEN_HEIGHT (240)
 
-#define ARRAYLENGTH(arr) (sizeof(arr)/sizeof(arr[0]))
+#define ARRAYLENGTH(arr) (s32)(sizeof(arr)/sizeof(arr[0]))
```
queueRAM, line 52: Needed this cast to avoid `sltiu` in `func_8002312C` in `soft_reset`. Didn't seem to impact other code, but wanted to call it out.

src/12DD0.c
```diff
@@ -1,17 +1,160 @@
...
+void func_800121D0(void) {
+    u16 index;
+    u16 actor_index;
 
-#pragma GLOBAL_ASM("asm/nonmatchings/12DD0/func_80012288.s")
+    actor_index = 0;
+    index = gActors[actor_index].health;
+    func_8001E2D0(actor_index);
+    gActors[actor_index].health = index;
+    gActors[0].posZ.whole = 1;
```
Bl00D4NGEL, line 11: question: is it not possible to use `gActors[actor_index]` here?

src/lifebar.c
```diff
@@ -1,3 +1,134 @@
...
+            if (gActiveFrames & 0x10) {
+                PALETTE_802209E0[index] = (((gActiveFrames & 0xF) << 12) + 0x801);
+            }
+            else {
+                PALETTE_802209E0[index] = (0xF801 - ((gActiveFrames & 0xF) << 12));
+            }
+        }
+    }
+    else if (gActors[0].health < 0) {
+        for (index = 0; index < 0x60; index++) { \
```
Bl00D4NGEL: question/suggest: what's the thought(s) on doing `for (index = 0; index < 0x60; index++)  PALETTE_802209E0[index] = 1;` here instead?

src/lifebar.c
```diff
@@ -1,3 +1,134 @@
...
+            PALETTE_802209E0[index] = 1;
+        }
+    }
+    else {
+        if (gHealthDisplayed < gActors[0].health) {
+            if (gHealthDisplayed + 5 < gActors[0].health) {
+                gHealthDisplayed = gHealthDisplayed + 5;
```
Bl00D4NGEL:
```suggestion
                gHealthDisplayed += 5;
```

src/lifebar.c
```diff
@@ -1,3 +1,134 @@
...
+            if (segment_1 < 0) {
+                segment_1 = 0;
+            }
+            if (segment_2 < segment_1) {
+                segment_2 = 0x60;
+            }
+        }
+        else if (gActors[0].health < gHealthDisplayed) {
+            if (gActors[0].health < gHealthDisplayed - 5) {
+                gHealthDisplayed = gHealthDisplayed - 5;
```
Bl00D4NGEL:
```suggestion
                gHealthDisplayed -= 5;
```

src/lifebar.c
```diff
@@ -1,3 +1,134 @@
...
+            }
+            if (segment_1 < 0) {
+                segment_1 = 0;
+            }
+            if (segment_2 < segment_1) {
+                segment_2 = 0x60;
+            }
+        }
+        else if (gActors[0].health < gHealthDisplayed) {
+            if (gActors[0].health < gHealthDisplayed - 5) {
```
Bl00D4NGEL: this might be "better" to read as a MIN/MAX `gHealthDisplayed = MIN(gActors[0].health < gHealthDisplayed - 5)` (whatever matches, if it does)

src/lifebar.c
```diff
@@ -1,3 +1,134 @@
...
+            PALETTE_802209E0[index] = ( ((D_800CBE20[index] & 0xF000) / 2) + ((D_800CBE20[index] & 0x780) / 2) + ((D_800CBE20[index] & 0x3C) / 2) + 1);
+        }
+        while (index < 0x60) {
+            PALETTE_802209E0[index++] = 1;
+        }
+    }
+    if (((prev_health_displayed + 0x3E7) / 0x3E8) < ((gHealthDisplayed + 0x3E7) / 0x3E8)) {
+        Sound_PlaySfx(0x141);
+    }
+    switch ((gHealthDisplayed - 1) / 0x3E8) {                            /* irregular */
```
Bl00D4NGEL:
```suggestion
    switch ((gHealthDisplayed - 1) / 0x3E8) {
```

src/12DD0.c
queueRAM: I wish I could find a way to use all `gActors[actor_index]` or `gActors[0]`, but unfortunately could only get this combination to match: https://decomp.me/scratch/I1I0E

src/lifebar.c
queueRAM: I'd certainly prefer to remove the `\`. @Drahsid do you have thoughts on formatting here? There is also a more egregious use of 4-line `\` in gameover.c

src/lifebar.c
queueRAM: Great idea! I changed both of these to use MIN/MAX

src/lifebar.c
Drahsid: Is the question:
```c
for (index = 0; index < 0x60; index++)  PALETTE_802209E0[index] = 1;
```
or if it matches:
```c
for (index = 0; index < 0x60; index++)  { PALETTE_802209E0[index] = 1; }
```
over the `\` syntax?

In this case, the matching requirement does tell us how the original code was formatted, so it's probably ok, though I prefer the explicit braces.

queueRAM: Question was intentionally open-ended since there could be more than three options on the table. :)
Thanks for your thoughts. I'll remove the `\` and go with the one with braces and see if I can apply this similarly to the one in gameover.c

src/gameover.c
```diff
@@ -1,9 +1,570 @@
...
+                actor_1 = 0;
+                if (gActors[actor_1].colorA < 0xFC) { \
+                    color_a = gActors[actor_1].colorA + 4; \
+                } \
+                else { \
+                    color_a = gActors[actor_1].colorA;
+                    if (color_a == 0xFC) {
+                        color_a = 0xFF;
+                    }
+                }
```
queueRAM, line 239: This is the multi-line `\`, but I can't figure out a cleaner way to implement this on a single-line given the multiple if blocks. Ternary doesn't seem to help: https://decomp.me/scratch/TLWw8

Drahsid: original code may have unironically been: `if (gActors[index].colorA < 0xFC) var_v0 = gActors[index].colorA + 4; else { var_v0 = gActors[index].colorA; if (var_v0 == 0xFC) var_v0 = 0xFF; }` all on one line.

src/12DD0.c
Drahsid: https://decomp.me/scratch/D3v8i

the 'if (1)' is not required but I suspect the original source was did something like this.

Another idea which I couldn't get to work was this being the result of an inlined function: https://decomp.me/scratch/uCl5i

queueRAM: I like that `if` block idea. This could also have been a `do { .. } while(0)` macro, but would be strange to use another variable like this: https://decomp.me/scratch/9bzkd
```c
#define ResetActor_RestoreHealth(actor_index) do { \
        u16 index = actor_index; \
        var_v1 = gActors[index].health; \
        func_8001E2D0(index); \
        gActors[index].health = var_v1; \
    } while (0)
```

Drahsid: do-while isn't needed; and the local-clobbering could be another macro argument: https://decomp.me/scratch/PdG9a
Note the `{}` are probably required if this appears anywhere that isn't at the start  of a function

Drahsid:
Is there anything holding this up? Do you need help with anything?

queueRAM:
> Is there anything holding this up? Do you need help with anything?

I rebased on master, resolved conflicts, and ensured progress is up-to-date. Let me know if there are any other changes you'd want, but otherwise I think it is ready.

src/gameover.c
queueRAM: I left this as-is with the `\` since I think it would be much more difficult to read on one-line

PR #41, opened 2026-07-10
Match more of 66250

Bl00D4NGEL:


include/cosine.h
```diff
@@ -10,11 +10,12 @@ extern f32 gCosineLookup[1024];
 extern f32 gUnusedFloatTable[512];
 
 #define COSLEN (ARRAYLENGTH(gCosineLookup) - 1)
-#define COSPiOver2 (ARRAYLENGTH(gCosineLookup) / 4)
-#define COSPiOver4 (ARRAYLENGTH(gCosineLookup) / 8)
+#define COSPiOver2 (ARRAYLENGTH(gCosineLookup) / 2)
```
queueRAM: Can you double-check the math here? I'm still not confident, but these were my notes when i reviewed earlier:
```
gCosineLookup[1024] // 0x400
COS(X) = gCosineLookup[X & 0x3FF]
SIN(X) = gCosineLookup[(X - 0x100) & 0x3FF]

sin(x) = cos(x - pi/2) // trig identity
→ pi/2 = 0x100
→ pi/2 = ARRAYLENGTH(table)/4
```

src/66250.c
```diff
@@ -954,29 +1154,194 @@ void func_80079760(u16 actor_index) {
...
+    u16 temp_v0;
+
+    gActors[actor_index].unk_128 -= 1.0f;
+    temp_v0 = Actor_RangeFindInactive(0x70, 0x7A);
+    if (temp_v0 != 0) {
+        ACTOR_INIT(temp_v0, 0);
+        gActors[temp_v0].posX.whole = gActors[actor_index].posX.whole;
+        gActors[temp_v0].posY.whole = gActors[actor_index].posY.whole;
+        gActors[temp_v0].posZ.whole = gActors[actor_index].posZ.whole - 1;
+        ((void (*)(u16, u16)) gActors[actor_index].var_154)(actor_index, temp_v0);
```
queueRAM: Can potentially use `Actor2Func` typedef here

src/66250.c
```diff
@@ -910,9 +1008,55 @@ void func_80078F14(u16 actor_index, s32 arg1) {
...
-#pragma GLOBAL_ASM("asm/nonmatchings/66250/func_80078F54.s")
+void func_80078F54(u16 actor_index) {
+    s32 var_v1;
+
+    gActors[actor_index].graphicFlags |= ACTOR_GFLAG_ROTZ;
+    var_v1 = gActors[actor_index].unk_168 / FIXED_UNIT(1);
+    if (gActors[actor_index].flags & ACTOR_FLAG_FLIPPED) {
+        var_v1 = (var_v1 + COSPiOver2) & COSLEN;
+    }
+    gActors[actor_index].rotateZ = var_v1 * 0.3515625;
```
queueRAM: Can use `RadStep` macro for 0.3515625. It's equal to `360/ARRAYLENGTH(gCosineLookup)`

include/28EF0.h
queueRAM, line 1: ❤️

queueRAM (approved):
Looks good. Just one question on trig changes and a couple suggestions for using macros.

include/cosine.h
Bl00D4NGEL: I honestly don't understand what you're trying to say here (mostly because my trigonomtry lessons have been.. ages ago at this point).

Are you saying the old names were / are correct? I made this change because it felt weird to ready `X / 4` as "xOver2" because X is divided by 4, not 2.

If the old names should say we should add a comment for future-us that looks at this again and thinks they're clever and rename it :D Also if they should stay I wonder what to call the current `COSPiOver2`? Just `COSPi`?

Drahsid: queue is right here, and this change should be rejected. The rename is semantically wrong:
- `pi / 2 (90 deg)` -> `1024 / 4` = `256`
- `pi / 4 (45 deg)` -> `1024 / 8` = `128`
- `pi (180 deg)` -> `1024 / 2` = `512`

So the old names correctly describe the angle in the operation, where the new name is plainly wrong. You're mistakenly describing the arithmetic, not the result. If you check any of the code using these you will see that there is now an angle discrepancy from how it reads to what it intends.

queueRAM: I see. After a brief discussion about this, I think we should keep the previous values `#define COSPiOver2 (ARRAYLENGTH(gCosineLookup) / 4)` in this PR to not hold it up. While it is accurate at 1/4th of the array of `gCosineLookup`, it is not very descriptive and we'll work out a more complete rename in a separate PR. Just looks like indexes and degrees are used in the code, so avoiding radians is probably the right move.

Bl00D4NGEL: Following / combining the conversation on Discord I've made the following changes:
- reverted my name changes (Sorry for the confusion on my end!)
- Renamed `COSPiOver2` to `COS_DEG_90`
- Renamed `COSPiOver4` to `COS_DEG_45`
- Added `COS_DEG_180` (which is `ARRAYLENGTH(gCosineLookup) / 2`)
- Renamed `RadStep` to `DegreesPerIndex`
- Renamed `COSLEN` to `COS_MASK`
- Added `DEG_TO_INDEX` and `INDEX_TO_DEG` and replaced usages accordingly

My only suggested change to these would be to rename `DegressPerIndex` to `DEG_PER_INDEX` to keep naming consistent, thoughts on that?

Drahsid: Makes sense to me

src/66250.c
Drahsid, line 1193: gActors[0] is the root player actor, there's a `gPlayerActor` define for readability. Is there a reason you reverted this?

Bl00D4NGEL: No reason other than me probably messing up merge conflict resolution, fixed now

Drahsid:
you good to merge this?

PR #42, opened 2026-07-10
map .data/.bss and split includes for src/1F1E0.c

Bl00D4NGEL:
- Moved all "dangling" protos in src/1F1E0.c into separate .h files with fitting names where applicable, otherwise default to <c-file-name>.h
- mapped bss / data
- Added comments for where unmapped & external data / bss belongs


src/1F1E0.c
```diff
@@ -194,232 +233,237 @@ void func_8001EC1C(void) {
...
-        sp198[0] = (gPlayerActor.posY.whole + gPlayerActor.hitboxBY1) - 0x28;
-        count_t0 = 1;
-        for (; index_s1 < 0x90; index_s1++) {
-            if (gActors[index_s1].flags & ACTOR_FLAG_UNK7) {
-                spBB8[count_t0] = index_s1;
-                sp978[count_t0] = gActors[index_s1].posX.whole + gActors[index_s1].hitboxAX1;
-                sp738[count_t0] = gActors[index_s1].posX.whole + gActors[index_s1].hitboxAX0;
-                sp3D8[count_t0] = gActors[index_s1].posY.whole + gActors[index_s1].hitboxAY0;
-                sp198[count_t0++] = gActors[index_s1].posY.whole + gActors[index_s1].hitboxAY1;
+    if ((D_80137458 & 0x10) || (gGameState != 6)) {
```
Bl00D4NGEL, line 1989: no actual changes here, just early-return to remove a level of nesting, does not affect :OK:

include/8D0A0.h
```diff
@@ -0,0 +1,10 @@
+#ifndef FILE_8D0A0_H
+#define FILE_8D0A0_H
+
+#include "common.h"
+
+// guessed signature
+void func_8008C528(u16);
```
queueRAM: I can help confirm signature is correct from PR#40

```suggestion
void func_8008C528(u16 portrait_index);
```

include/11820.h
```diff
@@ -0,0 +1,8 @@
+#ifndef FILE_11820_H
+#define FILE_11820_H
+
+#include "common.h"
```
queueRAM: I'm open to hear differing opinions, but I tend to prefer to only include what is actually used in the header. In this simple case, no headers are needed, but a default of `#include <PR/ultratypes.h>` is often enough. This helps avoid circular dependencies and marginally reduce compile times.

queueRAM (approved):
What a lovely PR. Thanks for sorting out the includes and 1F1E0 .data/.bss. Couple suggestions, but approving regardless.

include/11820.h
Bl00D4NGEL: That's valid, I totally forgot / ignored that this project's common.h contains *a bunch* of things. I'm used to common.h containing only the ultratype import as well as maybe some common macros (like MIN, MAX, CLAMP, maybe our COS macros) but this isn't the case here.

I replaced all the common.h with ultratypes.h includes. There's a few files that do not currently use the ultra types but I've kept the include in place to make moving things into that file just that little bit easier later on.

Drahsid: It might make more sense to trim down common.h to only common things, and I suppose move things into more relevant headers (ex: `DEFAULT_INT` to inttypes.h)

versions/us1/mischiefmakers.yaml
```diff
@@ -82,7 +82,7 @@ segments:
       - [0x001050, c, boot]
       - [0x001BE0, c, input]
       - [0x001DF0, c, dma]
-      - [0x001EF0, c, gamestate]
+      - [0x001EF0, c, gameState]
```
queueRAM: For consistency, I would prefer snake case `game_state` for file names

src/1F1E0.c
Drahsid, line 1: array formatting should be like:
```c
u16 abcxyz[] = {
    /* ... */ 
};

u16 cbazyx[] = {
    /* ... */ 
};
```
but it is more condensed in this file:
```
u16 abcxyz[] = {
/* ... */ };
u16 cbazyx[] = {
/* ... */ };
```

Note this only counts for multiline arrays. One-liners are fine.

versions/us1/mischiefmakers.yaml
Bl00D4NGEL: Should we also update floatTable and cosineTable?
They're currently also in camelCase
```yaml
      - [0x0BD8D0, .data, cosineTable]
      - [0x0BE8D0, .data, floatTable]

File: src/1F1E0.c
Comment (Bl00D4NGEL) [reply src/1F1E0.c:1]: I'm afraid I don't see this sort of formatting in this file / PR, could you give me an example of which variable might be offending this suggestion?

Comment (Drahsid) [reply src/1F1E0.c:1]: Maybe the github diff view is not displaying it right or a firefox bug

File: versions/us1/mischiefmakers.yaml (hunk 4 above)
Comment (blackgamma7) [reply versions/us1/mischiefmakers.yaml]: there seem to be equal filenames of camel_case and snakeCase. Do we have an established standard?

**Review by blackgamma7 — COMMENTED** [review]
was wondering about the standard for filenames:

File: versions/us1/mischiefmakers.yaml (hunk 4 above)
Comment (queueRAM) [reply versions/us1/mischiefmakers.yaml]: Good questions. Skimming the current splat config list, seems `snake_case` is more common for C source, but I'll defer to @Drahsid on the C filename convention.

File: src/1F1E0.c
Comment (queueRAM) [reply src/1F1E0.c:1]: I think this is one example on line 121 of 1F1E0.c where the entire declaration is on one line:
```c
u16 D_800CA2B8[] = { 0x004C, 0x0100, 0x0010, 0x8100, 0x000F, 0x0100, 0x0078, 0x8100, 0x0004, 0x0100, 0x0016, 0x0000, 0x000F, 0x8000, 0x0007, 0x8100, 0x0008, 0x0100, 0x0007, 0x0000, 0x001A, 0x8000, 0x000A, 0x0000, ... };
```
where recommendation is breaking it to multiple lines as (i've chosen 16-_bytes_ per line here):
```c
u16 D_800CA2B8[] = {
    0x004C, 0x0100, 0x0010, 0x8100, 0x000F, 0x0100, 0x0078, 0x8100,
    0x0004, 0x0100, 0x0016, 0x0000, 0x000F, 0x8000, 0x0007, 0x8100,
    0x0008, 0x0100, 0x0007, 0x0000, 0x001A, 0x8000, 0x000A, 0x0000,
...
};

Bl00D4NGEL: Ah I see, that does make sense then, I was looking for some multiline mishap, let me see if I can format those lines properly

Bl00D4NGEL: Looking at the usage it seems that these arrays were a tuple of `some id, button mask` so I went ahead and formatted them in pairs of 2 (so 4 bytes per line), this does inflate line count of the file but I don't think that's too much of an issue given we can just collapse those in our editors.

versions/us1/mischiefmakers.yaml
Drahsid: I don't really care, but I think either snake_case or PascalCase or Pascal_Snake_Case make more sense for file names than camelCase

PR #46, opened 2026-07-13
OK 26A00.c, bump python requirements and submodules

Drahsid:
Most of the changes here are whitespace and changing (u32) casts to (uintptr_t)


queueRAM (approved):
Looks good. Nice matches in 26A00!

PR #44, opened 2026-07-11
More labels, fix ALPHA_* index, add SFX_* indexes, "PlayerData" struct, complete gameover and stage, match OSD script.

blackgamma7:
- amend Alphabet.h to fix some indices and add others.
- add SFX_* indices, and replace magic numbers where applicable
- more various labels
- name "PlayerData" struct, fix fakematch addresses to struct.
- more ACTORTYPES for Festival Games.
- add .bss to stages
  - add appropriate `#define`'s to data
  - add sequential list of stages to header
- add .bss and .data to gameover
- add relevant struct and .bss to osd_debug_text
  - match some code, queueRAM has matched the rest.
- add various documentation
  - Add comments about Japanese differences
- . data for 42E90.c

PR #51, opened 2026-07-19
Rename and decompile 5A70 -> save_file

queueRAM:
- Decompile save_file and map .bss/.data/.rodata


src/save_file.c
```diff
@@ -0,0 +1,1454 @@
...
+            }
+        }
+        
+        if ((count == 1) || (!has_8FFF) ||
+            (D_80171ACC[SAVE_SLOT_0] >= 100) ||
+            (D_80171ACE[SAVE_SLOT_0] >= 2) ||
+            (D_80171AD0[SAVE_SLOT_0] >= 10000) ||
+            (D_80171AD4[SAVE_SLOT_0] >= 54) ||
+            (func_8001B244() != D_80171AD4[SAVE_SLOT_0]) ||
+            (D_80171AD8[SAVE_SLOT_0] >= 518400000) || // (60 * MSEC_PER_DAY) or (6000 * SEC_PER_DAY) ?
```
queueRAM: Does this `518400000` mean anything to anyone? It happens to equal `24 * 60 * 60 * 6000`

src/save_file.c
```diff
@@ -0,0 +1,1454 @@
...
+    0x014E, 0x014F, 0x0150, 0x0151, 0x0000, 0x00B8, 0x00B9, 0x00C5, 0x8FFF,
+};
+u16 D_800C5394[] = {
+    0x00A1, 0x00A2, 0x00A3, 0x00A4, 0x00A5, 0x00A6, 0x00A7, 0x00A8, 0x00A9,
+    0x00AA, 0x00BE, 0x00BC, 0x00BD, 0x0000, 0x00B1, 0x00B3, 0x00C6, 0x8FFF
+};
+
+// backed by EEPROM - put in common struct?
+//-----------------------------------
+u16 D_80171AA0[SAVE_SLOT_COUNT][11];
```
queueRAM: Any thoughts on the significance of this `11`?

Drahsid (approved):
Looks good, maybe fill out the SFX IDs from the enums/defines and make a macro/define for `(60 * MSEC_PER_DAY)` or `(6000 * SEC_PER_DAY)`

src/save_file.c
```diff
@@ -0,0 +1,1454 @@
...
+u16 D_800C5370[] = {
+    0x0145, 0x0146, 0x0147, 0x0148, 0x0149, 0x014A, 0x014B, 0x014C, 0x014D,
+    0x014E, 0x014F, 0x0150, 0x0151, 0x0000, 0x00B8, 0x00B9, 0x00C5, 0x8FFF,
+};
+u16 D_800C5394[] = {
+    0x00A1, 0x00A2, 0x00A3, 0x00A4, 0x00A5, 0x00A6, 0x00A7, 0x00A8, 0x00A9,
+    0x00AA, 0x00BE, 0x00BC, 0x00BD, 0x0000, 0x00B1, 0x00B3, 0x00C6, 0x8FFF
+};
+
+// backed by EEPROM - put in common struct?
```
queueRAM: `D_80171AA0` through `D_80171AD8` are only accessed individually in the code, except EEPROM access does a block transfer of 0x48 bytes together. Should I wrap these in a meta struct?

src/save_file.c
```diff
@@ -0,0 +1,1454 @@
...
+        else {
+            if (gFramesInScene % 2) { gActors[gActors[0xB7].unk_0A0 + 0].flags &= ~ACTOR_FLAG_DRAW; }
+            gActors[gActors[0xB7].unk_0A0 + 1].flags |= ACTOR_FLAG_DRAW;
+        }
+        if ((gButtonPress & gButton_Start) || (gButtonPress & gButton_A)) {
+            Sound_PlaySfx(0x23);
+            gActors[gActors[0xB7].unk_0A0 + 0].flags |= ACTOR_FLAG_DRAW;
+            gActors[gActors[0xB7].unk_0A0 + 1].flags |= ACTOR_FLAG_DRAW;
+            //-----------------------------------------------------------------
+            // something weird about indexes and instruction order here
```
queueRAM: oops, i shall remove this comment.

src/save_file.c
Drahsid: Maybe save name? 10 + null?

image

queueRAM: Good thought, i'll look into it

src/5A70.c
Drahsid, line 1: may be worth using `git mv` in the future

queueRAM: I did. According to my shell history, I used `git mv src/5A70.c src/save_file.c` but I also had to rename the `GLOBAL_ASM` paths so I think git just thought it was a new file in the end.

src/save_file.c
ThePlayerRolo: > Does this `518400000` mean anything to anyone? It happens to equal `24 * 60 * 60 * 6000`

this code kinda looks like checking for corrupted time save data?

Drahsid: I checked, it's the save file time.

queueRAM: #44 renamed this to `gFilePlayTimes`. do you know the units of this (seconds, frames, milliseconds)? does this mean maybe is a limit of 6000 days of play time perhaps for validation as suggested above?

src/save_file.c
queueRAM: This is now `gFileNames` and added `#define SAVE_SLOT_NAME_LENGTH 11`

src/save_file.c
Drahsid: I don't know the units, but setting it to max value reads as 99 days and minutes on the pause screen.

blackgamma7:
`D_800c4f20` was recently labeled `gEEPROMID`. looks to be a char[8].

queueRAM:
> `D_800c4f20` was recently labeled `gEEPROMID`. looks to be a char[8].

Thanks, working on updating this now. I'm also creating a `SaveFile` struct which encapsulates the gFileNames, gFileAges, etc. since they are bulk transferred from EEPROM.

src/save_file.c
queueRAM: After experimenting with wrapping the save file data in a struct, I observed the code no longer matches: https://decomp.me/scratch/DuOuJ
This means there is an implicit ordering of the data that is part of the 0x48 EEPROM transfers.

src/save_file.c
queueRAM: Added macro `#define FILE_PLAY_TIME_MAX (6000 * SEC_PER_DAY)`

queueRAM:
PR comments addressed, ready for review

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME,
+    DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME,
+    DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME,
+    DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME,
+    DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME,
+    DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME, DEFAULT_RECORD_TIME
+};
+
+u16 D_800C4FA8[] = { // save file name: "Start"
+    0x014A, 0x0131, 0x011E, 0x012F, 0x0131, 0x0000, 0x0000, 0x0000,
```
ThePlayerRolo: could this  be written with the '' style since its a string?

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    return (index + 1);
+}
+
+void func_80007578(void) {
+    Sound_PlaySfx2(SFX_MENU_BLIP);
+    gActors[0xB9].unk_0A0 = 0;
+    func_800073CC(0xC);
+}
+
+void func_800075A8(void) {
```
blackgamma7: I labeled this as `NameEntry_PrintKeyboardKATA` in the last PR.

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    u16 actor_index;
+
+    Sound_PlaySfx2(SFX_MENU_BLIP);
+    gActors[0xB9].unk_0A0 = 1;
+    actor_index = 0xC;
+    Text_InitActorGraphic(actor_index++, 0x11A, 0xFF80, 0x44, 0);
+    Text_InitActorGraphic(actor_index++, 0x11E, 0x80, 0x44, 0);
+    actor_index = Text_PrintStringRGB(actor_index, D_800C52A4, 0xFF90, 0x44, 0, 0x80, 0, 0);
+    actor_index = Text_PrintStringRGB(actor_index, D_800C52BC, 0x40, 0x44, 0, 0x80, 0, 0);
+    actor_index = Text_PrintString(actor_index, D_800C515C, 0xFF80, 0x30, 0);
```
blackgamma7: these arrays were labeled `gNameEntryRow#KATA` in the last PR

src/save_file.c
Drahsid: Not at the moment. Each character follows this game's own string encoding convention (see https://github.com/Drahsid/mischief-makers/blob/master/include/Alphabet.h)

We discussed this in a different PR, but at some point we might write a preprocessor to make it more readable. That said, @queueRAM should probably replace the literals here with alphabet macros

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    actor_index = Text_PrintString(actor_index, D_800C5180, 0xFF80, 0x20, 0);
+    actor_index = Text_PrintString(actor_index, D_800C51A4, 0xFF80, 0x10, 0);
+    actor_index = Text_PrintString(actor_index, D_800C51C8, 0xFF80, 0, 0);
+    Text_PrintString(actor_index, D_800C51EC, 0xFF80, 0xFFF0, 0);
+}
+
+void func_800076F0(void) {
```
blackgamma7: I labeled this `NameEntry_PrintKeyboardENG` last PR

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    gActors[actor_index++].flags = 0;
+    actor_index = Text_PrintStringRGBScale(actor_index, D_800C5304, 0xFF80,   0x30, 0, 0, 0, 0, 1.0f, 1.0f);
```
blackgamma7: these arrays were labeled `gNameEntryRow#ENG` in the last PR

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    u16 index;
+
+    Text_InitActorGraphic(actor_index++, 0x11E, 0x80, 0x44, 0);
+    gActors[actor_index++].flags = 0;
+    index = Text_PrintStringRGB(actor_index, D_800C52B0, 0x40, 0x44, 0, 0x80, 0, 0);
+    gActors[index++].flags = 0;
+    gActors[index++].flags = 0;
+    gActors[index++].flags = 0;
+    gActors[index++].flags = 0;
+    index = Text_PrintString(index, D_800C5084, 0xFF80,   0x30, 0);
```
blackgamma7: these arrays were labeled `gNameEntryRow#HIRA` in the last PR

src/save_file.c
ThePlayerRolo: > game's own str

ah I see

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+        gActors[0xBA].unk_0A0 = 1;
+        Sound_PlaySfx(SFX_MENU_DING);
+        gGameStateSubState++;
+    }
+    else {
+        Sound_PlaySfx(SFX_WRONG_0134);
+    }
+}
+
+u16 func_800073CC(u16 actor_index) {
```
blackgamma7: I labeled this as `NameEntry_PrintKeyboardHIRA` in the last PR.

src/save_file.c
queueRAM: Good question, but here the encoding of the string is using the `u16` codec defined in `Alphabet.h`. We are looking into a tool that could automatically convert these with some sort of pre-processor. e.g., here could become below and the pre-processor would convert to the expected u16 values.
```c
u16 D_800C4FA8[] = _("Start     ");
```

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+        gActors[0xB7].unk_0A0 + gActors[0xB8].unk_0A0,
+        (D_800C4FE8[gActors[0xB8].unk_0A0++] * 2) + 0x2D2,
+        (gActors[0xB1].unk_0A0 * 16) - 128,
+        (-gActors[0xB2].unk_0A0 * 16) + 48,
+        0x00, 0x40, 0x20, 0xFF);
+    gActors[gActors[0xB7].unk_0A0 + gActors[0xB8].unk_0A0].state = 0;
+    func_80007A90();
+}
+
+void func_80007BC0(u16* arg0, u16* arg1, u16* arg2) {
```
blackgamma7: `arg0` looks to be the selected Katakana char, likewise `arg1` for Katakana and `arg2` for English. likely why in `func_800078A4` , `gActors[0xB9].unk_0A0` is set to 2 in the English versions.

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    0x0000, 0x0000, 0x8FFF
+};
+
+u32 D_800C4FC0[] = {
+    0x00010000, 0x00020000, 0x00040000, 0x00000064,
+    0x00000003, 0x00020000, 0x00000003, 0x00040000,
+    0x00000000, 0x00040000,
+};
+
+u16 D_800C4FE8[] = {
```
blackgamma7: i had this labeled as `gNameEntrySpace`

src/save_file.c
blackgamma7: There are also 2 sets of English alphabet Characters, it also overwrites Katanana and Hiragana Characters in the English version for a third English alphabet face and some more symbols. Plus, we still only have a few of the Kanji characters ID'd.

src/save_file.c
```diff
@@ -0,0 +1,1476 @@
...
+    actor_index = Text_PrintStringRGB(actor_index, D_800C5078, 0xFFDC, 0x58, 0, 0x80, 0, 0x80);
+    actor_index += 3;
+    actor_index = func_800073CC(actor_index);
+    actor_index = Text_PrintStringRGB(actor_index, D_800C5138, 0x38, 0xFFE0, 0, 0x80, 0, 0);
+    actor_index = Text_PrintStringRGB(actor_index, D_800C513C, 0x58, 0xFFE0, 0, 0x80, 0, 0);
+    actor_index = Text_PrintStringRGB(actor_index, D_800C5140, 0x78, 0xFFE0, 0, 0x80, 0, 0);
+    actor_index = Text_PrintStringRGBScale(actor_index, D_800C5144, 0xFFB8, 0xFFC4, 0, 0, 0x60, 0, 1.0f, 1.0f);
+    Text_InitActorGraphicRGB(actor_index++, 0x464, 0xFFB8, 0xFFBC, 0, 0x60, 0xC0, 0);
+    gActors[0xB7].unk_0A0 = actor_index;
+    gActors[0xB9].unk_0A0 = gActors[0xB2].unk_0A0 = gActors[0xB1].unk_0A0 = 0;
```
blackgamma7: some suggestions on `#define`s  for these fields. 
`gActors[0xB9].unk_0A0` -> `NAME_ENTRY_LANGUAGE`
`gActors[0xB8].unk_0A0` -> `NAME_ENTRY_POSITION`
`gActors[0xB2].unk_0A0` -> `CHAR_SELECT_ROW`
`gActors[0xB1].unk_0A0` -> `CHAR_SELECT_COLUMN`

blackgamma7 (commented):
Had some suggestions based on my observations and the new labels:

src/save_file.c
queueRAM: To make sure I understand, do you mean add defines for the `0xB9` or for the entire data member or the whole thing?
```c
#define NAME_ENTRY_LANGUAGE 0xB9
gActors[NAME_ENTRY_LANGUAGE].unk_0A0 = 0;

#define NAME_ENTRY_LANGUAGE gActors[0xB9].unk_0A0
NAME_ENTRY_LANGUAGE = 0;
```

edit: i went with the latter `#define NAME_ENTRY_CHARSET  gActors[0xB9].unk_0A0` since i think that's what you meant, but please let me know otherwise

src/save_file.c
queueRAM: Thank you for catching these. I corrected these now and double-checked all save_file functions and data against the renames

src/save_file.c
blackgamma7: Yeah, this would be brief yet descriptive enough for me.

src/save_file.c
queueRAM: > replace the literals here with alphabet macros

Seems like a bit of work - does anyone already have a tool to assist with this? If not, I think my preference would be to replace the `0x8FFF` terminator with `ALPHA_NULL` in this PR and put the conversion work towards adding a more general text conversion tool.

src/save_file.c
```diff
@@ -0,0 +1,1488 @@
...
+#include "stage.h"
+#include "438E0.h"
+#include "82F80.h"
+
+#define MIN_PER_HOUR (60)
+#define SEC_PER_HOUR (60 * MIN_PER_HOUR)
+#define SEC_PER_DAY  (24 * SEC_PER_HOUR)
+#define MSEC_PER_DAY (100 * SEC_PER_DAY)
+
+#define DEFAULT_RECORD_TIME (10 * SEC_PER_HOUR)
```
blackgamma7: I also had this defined as `STAGE_MAX_TIME` in `stage.h`

src/save_file.c
```diff
@@ -0,0 +1,1488 @@
+#include "common.h"
+#include "save_file.h"
+#include "stage.h"
+#include "438E0.h"
+#include "82F80.h"
+
+#define MIN_PER_HOUR (60)
```
blackgamma7: could these be moved to `stage.h`? They could be expanded and used to define the target times for stages.

src/save_file.c
```diff
@@ -0,0 +1,1488 @@
...
+#define SEC_PER_HOUR (60 * MIN_PER_HOUR)
+#define SEC_PER_DAY  (24 * SEC_PER_HOUR)
+#define MSEC_PER_DAY (100 * SEC_PER_DAY)
+
+#define DEFAULT_RECORD_TIME (10 * SEC_PER_HOUR)
+#define FILE_PLAY_TIME_MAX (6000 * SEC_PER_DAY)
```
blackgamma7: This is also used in `Pause_PrintPlayTime` and `func_8000147C` (with a -1 offset)

blackgamma7 (commented):
a couple notes on the Time Macros

src/save_file.c
queueRAM: Thanks, I'll move these to `stage.h`. Also reviewing the stage code, I think I was mistaken on derivation of these units. These times are in units of ticks which is also what `gStageTimesToBeat[]` uses (e.g. 1080 ticks is 18 seconds)

src/save_file.c
Drahsid: `ALPHA_EN3_UPPER_S, ALPHA_EN3_LOWER_T, ALPHA_EN3_LOWER_A, ALPHA_EN3_LOWER_R, ALPHA_EN3_LOWER_T, ALPHA_SPACE, ALPHA_SPACE, ALPHA_SPACE`

queueRAM: There are many strings in this file. Does there exist a tool to help automate that conversion?

Drahsid: Not that I know of

blackgamma7: I can handle the conversion manually in another PR.

queueRAM: Thanks for the offer, but you don't need to do that. I can share the pain. I think I'll quickly put together something to do this macro conversion, but I would prefer to spend more of my time creating a tool to encode from utf-8 strings to the expected u16 encoding.

Drahsid: There are ambiguous command characters (different things in different contexts), and language conflicts, so such a tool would probably end up needing a bunch of iteration over time anyway

queueRAM: I had an idea to use a macro to show both string and contain values until we have a conversion tool when we can just drop the values. However, IDO doesn't support variadic macros and macros can't have commas so it gets a little ugly. Any ideas for improvements or other thoughts?
```c
#define SEP ,
#define _EN3(str, val) val

u16 gDefaultFileName[] = {
    _EN3("Start     ",
        ALPHA_EN3_UPPER_S SEP
        ALPHA_EN3_LOWER_T SEP
        ALPHA_EN3_LOWER_A SEP
        ALPHA_EN3_LOWER_R SEP
        ALPHA_EN3_LOWER_T SEP
        ALPHA_SPACE SEP
        ALPHA_SPACE SEP
        ALPHA_SPACE SEP
        ALPHA_SPACE SEP
        ALPHA_SPACE SEP
        ALPHA_NULL
    )
};
```

Drahsid: For the textconv-like tool, I was thinking making use of a script to do pre-processing and to have something like
```
u16 gDefaultFileName[] = _("[CSET:EN3]Start     ");
```
where game command characters and relevant text conversion preprocessing commands are in brackets or something (something like `[CSET:EN3]` only being useful if strings ever mix ambiguous characters, otherwise it could be something like `_EN3(...)`)

I was thinking a format like this would allow us to specify command characters in a readable way (ex: `_("[Festival:ColorYellow]Text is now yellow")` and to deal with unknown characters (ex: `_("[CHAR:0x7FFF]")`)

queueRAM: It seems there are a lot of considerations to make so it is unlikely we'll resolve this in this PR thread. For now, I have a script that parses Alphabet.h and converts u16 arrays to this format. I've updated my PR with this to show an example.
```c
u16 gDayEN2[] = { // _EN2(" day")
    ALPHA_SPACE,
    ALPHA_EN2_LOWER_D,
    ALPHA_EN2_LOWER_A,
    ALPHA_EN2_LOWER_Y,
    ALPHA_NULL
};
```

src/save_file.c
```diff
@@ -0,0 +1,1915 @@
...
+    ALPHA_EN3_UPPER_K,
+    ALPHA_EN3_UPPER_L,
+    ALPHA_EN3_UPPER_M,
+    ALPHA_SPACE,
+    ALPHA_PLUS,
+    ALPHA_DASH,
+    ALPHA_EQUALS,
+    ALPHA_NULL
+};
+u16 gNameEntryRow3ENG[] = { // _EN3("NOPQRSTUVWXYZ <><3")
```
queueRAM, line 563: Note: would propose use `♥` whenever we switch to a processing tool, but asm_processor doesn't appreciate this with the encoding of file
```
UnicodeEncodeError: 'euc_jp' codec can't encode character '\u2665' in position 53: illegal multibyte sequence
encoding with 'EUC-JP' codec failed
```

PR #57, opened 2026-07-26
Decompile 11820 and map data

queueRAM:


Drahsid (commented):
Looks fine at first glance. Will do a more thorough review after merge conflicts are solved.

queueRAM:
Conflicts resolved. I couldn't really figure out what was going on in the code, so there may be better matches. Please take your time to scrutinize this PR.

src/11820.c
```diff
@@ -1,30 +1,383 @@
...
+extern u16 D_80137474;
+extern u16 D_80137476;
+extern u16 D_80137478;
+extern u16 D_8013747C;
+extern u16 D_80137580[70];
+extern u16 D_80137610[70];
+
+extern u8* D_80137774;
+extern u8* D_8013777C;
+extern u8* D_80137784;
```
Drahsid, line 121: 26A00.c types these differently. They should be consistent.

src/11820.c
```diff
@@ -1,30 +1,383 @@
...
+    D_800C62A0, D_800C62A0, D_800C64A0, D_800C62A0,
+    D_800C64A0, D_800C64A0, D_800C62A0, D_800C60A0,
+    D_800C60A0, D_800C60A0, D_800C60A0, D_800C6CA0,
+    D_800C6020, D_800C64A0, D_800C6020, D_800C64A0,
+    D_800C62A0, D_800C60A0, D_800C60A0, D_800C6020,
+    D_800C6020, D_800C6FA0, D_800C6FA0, D_800C6FA0,
+    D_800C6FA0, D_800C75A0, D_800C76A0, D_800C6020,
+    D_800C76A0, D_800C76A0, D_800C76A0, D_800C76A0,
+    D_800C76A0, D_800C76A0, D_800C76A0, D_800C76A0,
+    0x00000000, 0x00000000, 0x00000000, 0x00000000
```
Drahsid: These zeroes mark the data split. This array has to do with maps/scenes, of which there are 88 entries (which is consistent with the rest of the game).

Drahsid (approved):
No real complaints.

src/11820.c
queueRAM: Good point on number of entries. I suspected data split and originally omitted the zeros, but then the data didn't line up.
```
0x000c88c0! 00000000000000000000000000000000    800128e0800128f080012944800128e0
0x000c88d0! 800128e0800128f080012944800128e0    800128e080012970800128e0800128e0
0x000c88e0! 800128e080012970800128e0800128e0    800128e0800129c880012a24800128e0
```

Alternatively, i could add another unused var afterwards assigned to zero and it will match again
```c
u8* D_800C7B60[] = {
...
    D_800C76A0, D_800C76A0, D_800C76A0, D_800C76A0
};

u32 unused = 0x0;
```

src/11820.c
queueRAM: Good catch. I aligned the 26A00.c types with these changes. I still can't quite figure out .bss mapping between the two files though.

src/11820.c
Drahsid: Oh, this is because the ending is already aligned, so you're right that there needs to be some data after this. I'm not sure there's another way to produce this, so the unused var might be the best bet.

PR #59, opened 2026-07-28
Use depfiles with ninja to rebuild C files on header changes

queueRAM:
- Use clang/cpp to generate *.o.d dependency files during cc phase
- When dependent headers change, ninja will rebuild C sources

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -520,7 +520,7 @@ def create_build_script(linker_entries: List[LinkerEntry]):
         "cc",
         description="cc $in",
         command=(
-            f"{CPP} {CPP_FLAGS} {COMMON_INCLUDES} {IDO_DEFS} -MD -MF $out.d -MT $out -o /dev/null $in"
+            f"{CPP} {CPP_FLAGS} {COMMON_INCLUDES} {IDO_DEFS} -MM -MF $out.d -MT $out $in"
             f" && {PYTHON} tools/asm_processor/build.py --input-enc=utf-8 --output-enc=EUC-JP --asm-prelude {ASM_PROCESSOR_PRELUDE} {IDO_CC} -- {CROSS_AS} {AS_FLAGS} -- -G 0 -non_shared -fullwarn {GAME_WARNING_SUPPRESSIONS} -verbose -Xcpluscomm -nostdinc -Wab,-r4300_mul $flags {MIPS_FLAGS} {COMMON_INCLUDES} {IDO_DEFS} -c -o $out $in"
         ),
         depfile="$out.d"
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -130,11 +130,6 @@ def preferred_tool(programs: List[str]) -> str:
 ASM_PROCESSOR_PRELUDE = "include/asm_processor_prelude.inc"
 GAME_WARNING_SUPPRESSIONS = "-woff 649,838"
 
-GAME_CC_CMD = f"{PYTHON} tools/asm_processor/build.py --input-enc=utf-8 --output-enc=EUC-JP --asm-prelude {ASM_PROCESSOR_PRELUDE} {IDO_CC} -- {CROSS_AS} {AS_FLAGS} -- -G 0 -non_shared -fullwarn {GAME_WARNING_SUPPRESSIONS} -verbose -Xcpluscomm -nostdinc -Wab,-r4300_mul $flags {MIPS_FLAGS} {COMMON_INCLUDES} {IDO_DEFS} -c -o $out $in"
```
queueRAM, line 133: I moved these inline with the ninja rules below. If you prefer to keep them here, can move them back.

configure.py
```diff
@@ -521,19 +513,23 @@ def build(
...
-        command=f"{LIBULTRA_AS_CMD}",
+        command=f"{IDO_CC} {LIBULTRA_ASFLAGS} {LIBULTRA_DEFAULT_MIPS_FLAGS} {LIBULTRA_DEFAULT_DEFS} {LIBULTRA_DEFAULT_ASOPT_FLAGS} $in {LIBULTRA_INCLUDES} -o $out && {O32_TOOL} $out && {CROSS_STRIP} $out -N asdasdasdasd"
     )
 
     ninja.rule(
         "cc",
         description="cc $in",
-        command=f"{GAME_CC_CMD}",
+        command=(
+            f"{CPP} {CPP_FLAGS} {COMMON_INCLUDES} {IDO_DEFS} -MD -MF $out.d -MT $out -o /dev/null $in"
```
queueRAM: This is the main line of the PR, using the `CPP` already in use to generate the dep files. It doesn't appear ninja allows running multiple commands in the same clean way as a makefile, so needed to chain with `&&`.

The only concern I have here is depending on `/dev/null` to redirect the CPP results to. This seems to work on macOS and Linux, but would certainly be unhappy on OSes like Windows. Couldn't find a way to instruct clang/cpp to just not output the preprocessed file.

Drahsid:
```suggestion
f"{CPP} {CPP_FLAGS} {COMMON_INCLUDES} {IDO_DEFS} -M -MF $out.d -MT $out $in"
```
isn't this what you want?

queueRAM: Why yes, that's much better. I tested it and it works just as I wanted on macOS/clang and Debian/gcc. I would appreciate someone else double-checking it doesn't disturb anything.

PR #69, opened 2026-08-14
OK func_800423A0, various cleanup

Drahsid:
This matches the last function in the main segment. We thought it required `gScreenPosCurrent[2]` over `gScreenPosCurrentX` and `gScreenPosCurrentY`, but `gScreenPosCurrent[2]` didn't seem to match in most instances.
I discovered that the following (adding `parts`) ended up working:
```c
typedef union {
    /* 0x00 */ s32 raw;
    /* 0x00 */ s16 parts[2]; // 0 = whole, 1 = frac
    struct {
    /* 0x00 */ s16 whole;
    /* 0x02 */ s16 frac;
    };
} FixedCoord;
```
This also helped solve and/or simplify some existing fakematches which weren't related to `gScreenPosCurrent...`.
I also updated Actor with an array at 170 which was needed for a good match; this also solved some really oddball address folding, which resulted in pointer-arithmetic fakematches in the past.

Finally, I noticed some fixed unit stuff around the code I was editing and cleaned that up, so there's some extra macro use.

### DIFF
--- a/include/actor.h
+++ b/include/actor.h
@@ -118,6 +118,8 @@ enum ActorFlags3 {
     ACTOR_FLAG3_UNK31 = (1U << 31U)
 };
 
+#define ACTOR_POSITION_HISTORY_COUNT 8
+
 typedef struct {
     /* 0x000 */ Mtx matrices[2]; // see A540: `actor + (gCurrentFramebufferIndex << 6)` before guTranslate/guScale/guRotate
     /* 0x080 */ u32 flags; // uses ActorFlags enum. 0 indicates inactive ("free") actor index
@@ -263,6 +265,7 @@ typedef struct {
                 /* 0x150 */ s32 var_150;
                 /* 0x150 */ s16 var_150_s16[2];
                 /* 0x150 */ s16* var_150_s16_ptr;
+                /* 0x150 */ FixedCoord unk_150; // TODO: array?
             };
             /* 0x154 */ s32 var_154;
             union {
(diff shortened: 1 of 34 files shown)


versions/us1/us1_report.json
```diff
@@ -50033,14 +50034,14 @@
       "id": "segment_main",
       "name": "Main segment",
       "measures": {
-        "fuzzy_match_percent": 99.85911,
+        "fuzzy_match_percent": 100.0,
```
queueRAM, line 50037: 💯

include/actor.h
```diff
@@ -292,47 +295,51 @@ typedef struct {
...
-                /* 0x184 */ s16 unk_184_s16[2];
-            };
-            union {
-                /* 0x188 */ s32 unk_188;
-                /* 0x188 */ s16 unk_188_s16;
-            };
-            union {
-                /* 0x18C */ s32 unk_18C; // field sometimes treated as int
-                /* 0x18C */ u16* palette_18C; // when ACTOR_GFLAG_PALETTE is set in graphicFlags, field is treated as palette pointer
+                /* 0x170 */ Position_U16 positionHistory[ACTOR_POSITION_HISTORY_COUNT]; // see: func_800423A0; TODO: find more uses
```
queueRAM: Totally on board with all these cleanups, looks so nice! One thing I didn't quite grasp from this `positionHistory` inclusion is the reason behind the `Position_U16` type. In my reading through the code, looked like all accesses were the `raw` and never the `positionX`/`positionY`. Is there something hiding that uses the `u16` fields?

queueRAM (approved):
Wow, so much nicer with all those cleanups and fakematch reduction. Nice work! Approving either way, but just had a question about the using the `Position_U16` type.

include/actor.h
Drahsid: `Position_U16` represents the actual data, so `s32` may actually be closer to the original source. This is basically forming whole-units (FixedUnit::whole) and packing two of them into a s32. See:
```c
position_x.raw = COS(gActors[actor_index].unk_16C) * FIXED_UNIT(4.5);
position_y.raw = SIN(gActors[actor_index].unk_16C) * FIXED_UNIT(4.5);
position_x.raw =
    FROM_FIXED((COS(angle) * FIXED_UNIT(40.0)) + position_x.raw) +
    gPlayerActor.posX.whole + gScreenPosCurrentX.parts[0];
position_y.raw =
    FROM_FIXED((SIN(angle) * FIXED_UNIT(40.0)) + position_y.raw) +
    gPlayerActor.posY.whole + gScreenPosCurrentY.parts[0];

position_x.raw = TO_FIXED(position_x.raw) + (u16)position_y.raw; // packing two into a 32 bit value here
```

The code which reads or interprets this is either currently written in some other form, not decompiled, or doesn't even exist in the rom.

queueRAM: Got it, thanks for the explanation. So it is packed X/Y data, but instead of using a struct or array, it's being packed in s32 with multiply and unpacked with a divide.
```c
position_x.raw = TO_FIXED(position_x.raw) + (u16)position_y.raw
```
and unpacked in the following code. it doesn't make sense to me that `posY` just uses `raw` directly without removing the upper `posX` component, but i guess that gets masked out from the implicit cast to 16-bit `posY.whole`?
```c
gActors[actor_index].posX.whole =
        FROM_FIXED(gActors[actor_index].positionHistory[ACTOR_POSITION_HISTORY_COUNT - 1].raw) -
        gScreenPosCurrentX.whole;
gActors[actor_index].posY.whole =
        gActors[actor_index].positionHistory[ACTOR_POSITION_HISTORY_COUNT - 1].raw -
        (u16)gScreenPosCurrentY.whole;
```

Drahsid: Yeah, though to be fair I'm pretty sure `<< 16` and `& 0xFFFF` etc. also match

queueRAM: Everything looks great to me. I stand by my approval

